PR #1, opened 2019-09-04
Feature/prod fixes

piotrjanosz:
- search overlay now appearing earlier (on input focus, not type)
- display ‘no results for x’ when no search suggestions matching user query
- added link to docs in mobile dropdown menu
- fixed svg issue for mobile safari
- fixed a glitch on mobile docs where filters for guides / tools would not initially show up
- some performance gains for autocomplete by moving functions that do not need to be rerendered outside of the component


chriskalani:
Updates to search look great!

fabioberger (approved):

PR #3, opened 2019-09-04
Update reference to getTransactionHashHex

xianny:
Was previously in @0x/contract-wrappers.transactionEncoder, which has been removed.
Redirect users to @0x/order-utils.transactionHashUtils

### DIFF
--- a/mdx/guides/0x-extensions-explained.mdx
+++ b/mdx/guides/0x-extensions-explained.mdx
@@ -102,20 +102,23 @@ require(isWhitelisted[takerAddress], 'TAKER_NOT_WHITELISTED');
 EXCHANGE.executeTransaction(salt, signerAddress, signedExchangeTransaction, signature);
 ```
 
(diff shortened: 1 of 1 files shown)


mdx/guides/0x-extensions-explained.mdx
```diff
@@ -102,20 +102,23 @@ require(isWhitelisted[takerAddress], 'TAKER_NOT_WHITELISTED');
...
 ```
 
-In order to implement an on-chain whitelist, the contract would first query if the user is present in a list, if the user is present then the 0x transaction is executed. Our contract wrappers package exposes a [transaction encoder](https://0x.org/docs/tools/contract-wrappers#TransactionEncoder) to encode 0x transactions. An detailed example of encoding and executing 0x transactions can be found in our [0x-starter-project](https://github.com/0xProject/0x-starter-project/blob/master/src/scenarios/execute_transaction.ts#L20).
+In order to implement an on-chain whitelist, the contract would first query if the user is present in a list, if the user is present then the 0x transaction is executed. `@0x/order-utils` exposes [transaction hash utils](https://0x.org/docs/tools/order-utils/v8.3.1#const-transactionhashutils) to help encode 0x transactions. When encoded and signed, a 0x transaction can be executed by any third party such as a relayer or extension contract. An detailed example of encoding and executing 0x transactions can be found in our [0x-starter-project](https://github.com/0xProject/0x-starter-project/blob/master/src/scenarios/execute_transaction.ts#L20).
```
xianny, line 105: the `0x-starter-project` link is updated, thanks to @dekz 🙏

fabioberger (approved):
Thanks Xianny!

PR #5, opened 2019-09-09
[WIP] Added fetching tools from S3 and rendering them with runtime mdx

fabioberger:


ts/pages/docs/page.tsx
```diff
@@ -33,13 +33,22 @@ import { IContents, TableOfContents } from 'ts/components/docs/sidebar/table_of_
...
 
 import meta from 'ts/utils/algolia_meta.json';
 
+// @ts-ignore no type definition for this one
+import MDXRuntime from '@mdx-js/runtime';
+const remark = require('remark');
+const remarkSlug = require('remark-slug');
+const remarkAutolinkHeadings = require('../../../webpack/remark_autolink_headings');
+const remarkSectionizeHeadings = require('../../../webpack/remark_sectionize_headings');
+const remarkTableOfContents = require('../../../webpack/remark_table_of_contents');
```
fabioberger, line 47: Now that these are not only used in `webpack`, let's move them to a differently named dir.

webpack.config.js
```diff
@@ -188,5 +188,9 @@ module.exports = (_env, argv) => {
     config.plugins = plugins;
     console.log('i ｢atl｣: Plugin Count: ', config.plugins.length);
 
+    config.node = {
```
fabioberger, line 191: Can you add a comment above this explaining why it's necessary.

fabioberger:
Remaining TODO's:
    - Rendering 0x.js docs is broken -- seems too big. Causes CPU to spike to 200%. Times out.
    - Populate version dropdown from S3 versions _not_ algolia_meta.json

PR #8, opened 2019-09-15
Webpack optimization

kimpers:
This makes some low hanging optimizations to the Webpack configuration for faster builds.

## Optimizations (dev build):

- `useCache`: Use internal file cache when building. ~ 5 - 10s improvement
- `transpileOnly`: Don't check TS types when building in Webpack for development. When developing you normally have Typescript checking inside your editor so there is no reason to also check in Webpack. ~ 10 - 15s improvement. To make sure that type checking is run now that it's not part of the Webpack build I also added it as a pre-push hook with [husky](https://github.com/typicode/husky).
-  `devtool`: [cheap-module-eval-source-map](https://webpack.js.org/configuration/devtool/). Slightly worse source maps for local dev, it only maps line numbers and not columns. ~ 5s improvement.

## Build times
Dev: Initial build after clean clone ~ 55s
Dev: Subsequent builds ~ 25s
Prod: ~ 200s

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
         "clean": "shx rm -rf public/bundle*",
         "lint": "tslint --format stylish --project . 'ts/**/*.ts' 'ts/**/*.tsx'",
         "fix": "tslint --fix --format stylish --project . 'ts/**/*.ts' 'ts/**/*.tsx'",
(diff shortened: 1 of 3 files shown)


fabioberger (approved):
Nice, LGTM

PR #11, opened 2019-09-21
Staking pool detail row

kimpers:
Staking pool overview details row. Data that this component will receive is still a bit unclear for me but this can be iterated upon when hooking this up to the general page. 

**Small device**
image

**Medium device**
image

**Desktop**
image

### DIFF
--- a/ts/components/staking/staking_pool_detail_row.tsx
+++ b/ts/components/staking/staking_pool_detail_row.tsx
@@ -18,7 +18,7 @@ const DesktopOnlyWrapper = styled.div`
     ${desktopOnlyStyle};
 `;
 
-const StakingPoolDetailsWrapper = styled.div`
+const StakingPoolDetailRowWrapper = styled.div`
     min-height: 120px;
     border: 1px solid #d9d9d9;
     display: flex;
@@ -157,7 +157,7 @@ const PoolWebsiteLink = ({ websiteUrl }: { websiteUrl: string }) => (
     </a>
 );
 
-interface IStakingPoolDetailsProps {
+interface IStakingPoolDetailRowProps {
     name: string;
     ethAddress: string;
     feesCollectedEth: BigNumber;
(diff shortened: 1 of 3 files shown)


ts/components/staking/staking_pool_detail_row.tsx
```diff
@@ -0,0 +1,218 @@
...
+            stroke-linejoin="round"
+        />
+        <path
+            d="M6.75 12.25L10.5 16L18 8.5"
+            stroke="#00AE99"
+            stroke-miterlimit="10"
+            stroke-linecap="round"
+            stroke-linejoin="round"
+        />
+    </svg>
```
kimpers: Could not find any matching check mark already in the codebase so I copied straight from the design. Worth generalizing?

dave4506: Not that I officially know if we have one, but I think we may have used this check mark for bullet lists, maybe it is in one of the landing pages.

kimpers: Sweet, I'll do some digging tomorrow. Thanks 👍

chriskalani: We do have one (bottom of page: https://0x.org/launch-kit), but this one w/ the circle is different. We should use both.

rensverschuren:
Are you using TSLint? Got some "semi-colon expected". Otherwise looking really good! I’ll add the semi-colons (just a few keystrokes away).

kimpers:
> Are you using TSLint? Got some "semi-colon expected". Otherwise looking really good! I’ll add the semi-colons (just a few keystrokes away).

Whoops 😳I do have TSLint setup but it does not seem to lint inside styled components. Will look into how to correct that. Thanks for fixing it 👍

fabioberger:
@kimpers TSLint not linting within styled components is a known issue. Not sure there is a work-around.

PR #12, opened 2019-09-24
Basic stylelint linting for styled-components

kimpers:
I noticed that TSLint did not catch some missing `;` in css inside a styled-component and that got me thinking that it would be nice to have linting on CSS as well.

This is a suggestion to add very basic styled-component aware css linting. It comes with some caveats though, like you can't do `//` style comments inside of a styled-component as they are not valid CSS.

This is the bare minimum, more rules can be added but that requires a bigger effort of cleaning up the styles.

Would this be something useful or not? Wdyt?


kimpers:
Closing this for now, will look into better solutions in the future

PR #15, opened 2019-09-27
Add trading pair types

rensverschuren:
Hi David,

I’ve set up a basic type for the trading pairs as shown on the Market Maker Profile page: 

Screenshot 2019-09-27 at 11 56 00

I will update this pull request with more types, but this way we can agree on a format ;-)

### DIFF
--- a/ts/types.ts
+++ b/ts/types.ts
@@ -731,6 +731,24 @@ export interface WebsiteBackendTradingPairs {
     tradingPairs: WebsiteBackendTradingPair[];
 }
 
+export interface StakingPoolMetrics {
+    totalVolume: string;
+    totalStaked: string;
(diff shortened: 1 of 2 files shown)


dave4506:
@rensverschuren Hey, structurally things look good, with minor details to be changed per making the backend client more generalized for 3rd party usage. 

The key changes I will be requesting/making will be changing the keys in the interface to align them more with Ethereum web3 and 0x terminology, will be making a commit per that on this branch per those changes (or I can just suggest them all to you through a review, and you can roll them in as you work on the front end)

rensverschuren:
Sounds good! If you can commit the changes to this branch, I’ll check if it works with the frontend.

PR #26, opened 2019-10-10
Use correct logo for VeriDex on explore page

kimpers:
image

Correct logo from the Verisafe team, unfortunately this is pretty large at 172KB due to the embedded png.


chriskalani:
I'd rather use the glyph SVG we had before @kimpers - We use simplified logos on Explore. Gods Unchained is a good example of how we simplified their logo

chriskalani:
Let's abandon this

kimpers:
> I'd rather use the glyph SVG we had before @kimpers - We use simplified logos on Explore. Gods Unchained is a good example of how we simplified their logo

Sounds good 👍 closing this

PR #42, opened 2019-10-24
Connect wallet dialog redux

kimpers:
This hooks up the connect wallet button from the menu with the connect wallet dialog. Will go through and add other "connect wallet" buttons as well.

### DIFF
--- a/ts/components/staking/layout/staking_page_layout.tsx
+++ b/ts/components/staking/layout/staking_page_layout.tsx
@@ -7,6 +7,8 @@ import { DocumentTitle } from 'ts/components/document_title';
 
 import { documentConstants } from 'ts/utils/document_meta_constants';
 
+import { ConnectWalletDialog } from 'ts/containers/connect_wallet_dialog';
+
 interface IStakingPageLayoutProps {
     children: React.ReactNode;
     title: string;
@@ -29,6 +31,7 @@ export const StakingPageLayout: React.FC<IStakingPageLayoutProps> = props => {
             />
 
             {props.children}
+            <ConnectWalletDialog />
         </SiteWrap>
     );
 };
--- a/ts/components/staking/layout/staking_page_layout.tsx
+++ b/ts/components/staking/layout/staking_page_layout.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { SiteWrap } from 'ts/components/siteWrap';
-import { Header as StakingHeader } from 'ts/components/staking/header/header';
+import { Header as StakingHeader } from 'ts/containers/staking/header/header';
 
 import { DocumentTitle } from 'ts/components/document_title';
 
--- a/ts/containers/connect_wallet_dialog.ts
+++ b/ts/containers/connect_wallet_dialog.ts
@@ -1,0 +1,31 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Dispatch } from 'redux';
+import { State } from 'ts/redux/reducer';
+
+import { ConnectWalletDialog as ConnectWalletDialogComponent } from 'ts/components/dialogs/connect_wallet_dialog';
+import { Dispatcher } from 'ts/redux/dispatcher';
+
+interface IConnectWalletDialogProps {}
+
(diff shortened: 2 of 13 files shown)


ts/components/staking/layout/staking_page_layout.tsx
```diff
@@ -29,6 +31,7 @@ export const StakingPageLayout: React.FC<IStakingPageLayoutProps> = props => {
             />
 
             {props.children}
+            <ConnectWalletDialog />
```
kimpers, line 34: Not sure if this is the best place to put the dialog component but here it is ensured to always be mounted on a page where a "Connect Wallet" button can exist. Wdyt?

ts/containers/connect_wallet_dialog.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { ConnectWalletDialog as ConnectWalletDialogComponent } from 'ts/components/dialogs/connect_wallet_dialog';
+import { Dispatcher } from 'ts/redux/dispatcher';
+
+interface IConnectWalletDialogProps {}
+
+// TODO(kimpers): add blockchain integration
+interface IConnectedState {
+    isOpen: boolean;
+}
```
kimpers, line 14: More state to come here regarding the actual connection

ts/containers/staking/header/header.ts
```diff
@@ -0,0 +1,31 @@
...
+    location?: Location;
+    isNavToggled?: boolean;
+    toggleMobileNav?: () => void;
+}
+
+interface IConnectedDispatch {
+    dispatcher: Dispatcher;
+}
+
+interface ConnectedState {}
```
kimpers, line 19: In the future this will need to know if a wallet is connected for the indicator on the mobile menu

ts/containers/connect_wallet_dialog.ts
```diff
@@ -0,0 +1,31 @@
...
+    dispatcher: Dispatcher;
+}
+
+const mapStateToProps = (state: State, _ownProps: IConnectWalletDialogProps): IConnectedState => ({
+    isOpen: state.isConnectWalletDialogOpen,
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<State>): IConnectedDispatch => ({
+    dispatcher: new Dispatcher(dispatch),
+});
```
fragosti: I know you see this in the code-base a lot, but I think it is best that we don't pass `Dispatcher` in as a prop when possible. Components should implement callbacks like "onDismiss" and the container code should worry about what to dispatch and when. Basically if component code doesn't contain the term "dispatch" that's good.

ts/components/staking/layout/staking_page_layout.tsx
fragosti: This makes sense to me.

ts/containers/connect_wallet_dialog.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface IConnectedDispatch {
+    dispatcher: Dispatcher;
+}
+
+const mapStateToProps = (state: State, _ownProps: IConnectWalletDialogProps): IConnectedState => ({
+    isOpen: state.isConnectWalletDialogOpen,
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<State>): IConnectedDispatch => ({
```
fragosti: FWIW I don't know when we started using the capital "I" `Interface` convention but it is against our style-guide unfortunately. Not sure if it's worth changing all the code now, but new code should adhere to the convention. Surprised that tslint isn't picking this up.

fragosti (approved):

ts/containers/connect_wallet_dialog.ts
kimpers: Oh, that might be my bad actually. I'm so used to prefixing them with "I". I'll fix here and the clean up as I see them in the code base in the future.

ts/containers/connect_wallet_dialog.ts
kimpers: Makes total sense to keep dispatch contained to the containers, will fix and keep this in mind for the future 👍

PR #43, opened 2019-10-30
Fixed spacing in governance

dave4506:
Screen Shot 2019-10-30 at 6 40 13 PM


chriskalani:
For the ZEIP rows, the padding on the top and bottom is still too big. And we lost the appropriate padding on the left and right.

kimpers:
I think #45 fixes the underlying issue here

PR #45, opened 2019-10-31
Fix Governance page styling and Section component props being applied twice

kimpers:
I think this fixes the underlying issue which was that the styling being passed to the Section component unintentionally was being passed to the inner wrapper component as well. The lead to margin and paddings begin applied twice.

@chriskalani there also is a rather large margin of `150px` being added after the vote cards section. Should that remain?

image


dave4506 (approved):
This is nice; I feel like we should have done this from the beginning; thanks @kimpers for adding this.

PR #46, opened 2019-11-01
Fix ZEIP vote card description layout

kimpers:


ts/components/newLayout.tsx
```diff
@@ -27,16 +28,13 @@ export interface WrapStickyProps {
 
 export interface SectionProps extends WrapProps {
     isPadded?: boolean;
-    isFullWidth?: boolean;
-    isFlex?: boolean;
```
kimpers, line 31: These are inherited from `WrapProps`

ts/components/newLayout.tsx
```diff
@@ -14,6 +14,7 @@ export interface WrapProps {
     isTextCentered?: boolean;
     isCentered?: boolean;
     isWrapped?: boolean;
+    isFlex?: boolean;
```
kimpers, line 17: Wrap was expecting this to get passed

dekz (approved):
LGTM

PR #49, opened 2019-11-05
Staking blockchain connect

kimpers:


ts/utils/web3_connector.ts
```diff
@@ -0,0 +1,224 @@
+import { getContractAddressesForNetworkOrThrow } from '@0x/contract-addresses';
```
kimpers: This was mostly copied from the Governance [VoteForm](https://github.com/0xProject/website/blob/development/ts/pages/governance/vote_form.tsx) component and refactored into an isolated Web3 utility.

ts/redux/sagas.ts
```diff
@@ -0,0 +1,23 @@
+import { call, put, takeLatest } from 'redux-saga/effects';
```
kimpers: I know it isn't ideal to add another dependency but redux does not handle async actions very well. 

I think this is a pretty clean way to expose slow async blockchain tasks globally through redux state and actions.

ts/components/dialogs/connect_wallet_dialog.tsx
```diff
@@ -293,7 +294,9 @@ export const ConnectWalletDialog = ({ isOpen, onDismiss }: ConnectWalletDialogPr
         walletProviders = [
             {
                 title: 'Detected wallet',
-                providers: [{ name: 'MetaMask', id: Providers.Metamask }],
+                providers: [
+                    { name: 'MetaMask', id: Providers.Metamask, onClick: () => onConnectWallet(Providers.Metamask) },
+                ],
```
kimpers: This still needs to be cleaned up.

yarn.lock
```diff
@@ -9940,6 +9998,14 @@ redux@^3.6.0:
     loose-envify "^1.1.0"
     symbol-observable "^1.0.3"
 
+redux@^4.0.4:
```
kimpers: If we go with this implementation I will update our Redux to be in line with this so that we don't need multiple versions.

ts/containers/connect_wallet_dialog.ts
```diff
@@ -28,6 +32,9 @@ const mapDispatchToProps = (dispatch: Dispatch<State>): ConnectedDispatch => {
         onDismiss: (): void => {
             dispatcher.updateIsConnectWalletDialogOpen(false);
         },
+        onConnectWallet: (provider: Providers): void => {
+            dispatcher.connectWallet(provider);
```
fragosti: So while I have heard good things about redux-saga, I am trying to understand the benefits over async / await, which is what we have leaned on heavily in other instances. Without redux-saga, we would just make this function async, await and dispatch the various things. What additional things are we getting here?

ts/utils/web3_connector.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    private _getNameGivenProvider(provider: ZeroExProvider): string {
+        const providerType = utils.getProviderType(provider);
+        const providerNameIfExists = providerToName[providerType];
+        if (providerNameIfExists === undefined) {
+            return constants.PROVIDER_NAME_GENERIC;
+        }
+        return providerNameIfExists;
+    }
+    private async _getProviderAsync(
```
fragosti: This code might be helpful. https://github.com/0xProject/0x-monorepo/blob/development/packages/instant/src/util/provider_factory.ts 

as well as this code.

https://github.com/0xProject/0x-monorepo/blob/development/packages/instant/src/util/provider_state_factory.ts

Potential alternative architecture:
I would love to have a similar structure to Instant. Ie. we store a lot of the web3 info in the redux state, and do not rely on initializing this class to get access to it.
When we do need to make an RPC call, I think it would be best to lean on web3Wrapper (which would be in the redux state) and pass it into a utility function.

fragosti (commented):
Looks good. Just need to be bought into sagas a bit more. Also pointed you to some instant code, which is generally way better than the website code.

ts/utils/web3_connector.ts
kimpers: I really like this approach! I did not realize that this existed 😳 

I will refactor to use these factories. Longer term it might even make sense to move this into module that can be shared between instant and the website. 👍

ts/containers/connect_wallet_dialog.ts
kimpers: The main advantages with sagas is that you get to model your async side effects into isolated worker tasks. They react to actions and can trigger state changes as well as new actions. The components just have to worry about triggering an action and the saga will take care of handling the side effect(s) of that action.

We could just make the container method async let that handle the async processing and dispatching the appropriate response and if that is an established pattern in other places it may be preferable.

The slight downside of this would be:
- the component now needs to be aware of the asynchronous nature of the dispatcher function and need to handle that appropriately
- it may get trickier to combine multiple side effects. In Sagas one saga can trigger multiple other sagas and so on without the component that triggered the action having to worry about it
- Sagas offer a simple way to debounce long running actions. `takeLatest` will ignore additional action triggers while the saga responding to it is already executing. This may need to be handled manually with async dispatchers.

I reached for Sagas because that is how I have previously handled async side effects in a Redux environment. I checked for existing patterns for handling this in the codebase (but only the website) and didn't find any so I did not know there was a precedence available.

I don't really have any strong opinions about how to handle this, if you think it worked well the way you've handled it in instant I'd be happy to follow that pattern 👍 Like I said previously, it's been a while since I worked with Redux, the status quo may have changed since 😄

fragosti: I'm comfortable leaving this up to you, and you'll be doing most of the work with it in the short-medium term.

To your points (respectively):
- I'm not sure I'm understanding why the component needs to be aware of the async nature of the dispatch. Ideally the component would not know if anything is being dispatched.
- "In Sagas one saga can trigger multiple other sagas and so on without the component that triggered the action having to worry about it" -- I think this is the main selling point. In instant we don't run into too much async complexity. Most actions are of the form "I'm doing some async things" ->"Do the async things" -> "Done with my list of async things". I think for this, async/await is enough. You can share logic with an async function and it is easy to understand. Do you think we will be composing async actions a lot? I think that is the real question, if not then async/await is enough IMHO.
- Yes this makes sense, but if we don't have async actions triggering async actions, then "debounce" is enough for straight up handlers.

kimpers: Let's try with async dispatcher functions. Nice to have consistency between the projects and it's easy enough move to Sagas later on if needed. I'm all for minimising external dependency reliance 😄

ts/utils/providers/constants.ts
```diff
@@ -0,0 +1,27 @@
+import { AccountNotReady, AccountState } from 'ts/types';
```
kimpers: TODO: keep separate or merge with main website?

ts/utils/providers/types.ts
```diff
@@ -0,0 +1,10 @@
+export enum ProviderType {
```
kimpers: TODO: rename and merge with website types

ts/containers/connect_wallet_dialog.ts
```diff
@@ -28,6 +37,21 @@ const mapDispatchToProps = (dispatch: Dispatch<State>): ConnectedDispatch => {
...
+                const providerState = providerStateFactory.getInitialProviderState(fallbackNetworkId);
+
+                const networkId = await providerState.web3Wrapper.getNetworkIdAsync();
+                if (networkId !== fallbackNetworkId) {
+                    dispatcher.updateNetworkId(networkId);
+                }
+
+                await asyncDispatcher.fetchAccountInfoAndDispatchToStore(providerState, dispatcher, true);
+            } catch (err) {
+                // TODO(kimpers): handle errors
```
kimpers: We need a way to handle error messages. There is a SnackBar component already in the project that we could utilize. Wdyt?

fragosti (approved):

PR #62, opened 2019-11-20
Query 0x API and render pools data

fragosti:
Renders the response found in https://github.com/0xProject/0x-api/pull/23

We do not have metadata but it renders correctly.

Screen Shot 2019-11-19 at 7 02 34 PM

I also upgraded us to the most recent versions of `redux` and `react-redux`


ts/pages/staking/home.tsx
```diff
@@ -10,44 +11,23 @@ import { StakingConfirmationDialog } from 'ts/components/dialogs/staking_confirm
...
     const [isStakingConfirmationOpen, toggleStakingConfirmation] = React.useState(false);
+    const [stakingPools, setStakingPools] = React.useState<PoolWithStats[] | undefined>(undefined);
+    React.useEffect(() => {
+        const fetchAndSetPoolsAsync = async () => {
+            const resp = await apiClient.getStakingPoolsAsync();
+            setStakingPools(resp.stakingPools);
+        };
+        // tslint:disable-next-line:no-floating-promises
+        fetchAndSetPoolsAsync();
+    });
```
kimpers: Correct me if I'm wrong but `useEffect` with no second arg would trigger on every component render and re-fetch the data. We would rather fetch it only once on page load right?
```suggestion
    }, []);
```

ts/utils/configs.ts
```diff
@@ -8,6 +8,9 @@ export const configs = {
     BACKEND_BASE_PROD_URL: 'https://website-api.0x.org',
     BACKEND_BASE_STAGING_URL: 'https://staging-website-api.0x.org',
     BACKEND_BASE_DEV_URL: 'https://localhost:3001',
+    API_BASE_PROD_URL: 'https://api.0x.org',
+    API_BASE_STAGING_URL: 'https://staging.api.0x.org',
+    API_BASE_DEV_URL: 'http://localhost:3000',
```
kimpers, line 14: NIT: I think `website-backend` defaults to using port `3000` for http, would be nice if they don't conflict when running locally.

kimpers (changes requested):

ts/pages/staking/home.tsx
fragosti: Oops!

ts/utils/configs.ts
fragosti: In here `BACKEND_BASE_DEV_URL` is using port 3001 already so they do not conflict.

package.json
```diff
@@ -77,7 +77,7 @@
         "react-tabs": "^3.0.0",
         "react-tooltip": "^3.2.7",
         "react-transition-group": "^4.2.1",
-        "redux": "^3.6.0",
+        "redux": "^4.0.4",
```
kimpers, line 80: 🎉

kimpers (approved):

PR #64, opened 2019-11-24
Combine staking wizard changes

kimpers:
There was quite a bit of overlap in our branches so I tried to cherry pick the differences from my branch into yours


ts/containers/staking/wizard/wizard.ts
```diff
@@ -28,10 +37,26 @@ const mapDispatchToProps = (dispatch: Dispatch<Action>): ConnectedDispatch => {
...
+            const currentAllowance = account.zrxAllowanceBaseUnitAmount || new BigNumber(0);
+
+            if (amountToStakeBaseUnits.isGreaterThan(currentAllowance)) {
+                await asyncDispatcher.increaseZrxAllowanceAndDispatchToStoreIfNeededAsync(
+                    providerState,
+                    networkId,
+                    amountToStakeBaseUnits,
+                    dispatcher,
+                );
+            }
```
kimpers, line 61: In order to get the staking contract interactions working it seems like I need to upgrade the `0x` dependencies to the latest `beta` versions and this leads to a lot of errors due to api and type changes in packages. 

I will proceed with the upgrade in a separate branch as this will probably be quite significant changes just to get everything up and running again with the latest modules. After that I can circle back to actually hooking things up with the contract.

ts/components/staking/wizard/wizard_flow.tsx
```diff
@@ -312,20 +313,25 @@ export const WizardFlow: React.FC<WizardFlowProps> = props => {
                 topLabels={['Amount', `Available: ${formattedAmount} ZRX`]}
                 labels={[StakingPercentageValue.Fourth, StakingPercentageValue.Half, StakingPercentageValue.All]}
                 value={stakeAmount}
+                selectedLabel={selectedLabel}
```
fragosti, line 316: Ah didn't realize this was a thing. When you type does it reset to no selected label value?

ts/containers/staking/wizard/wizard.ts
```diff
@@ -28,10 +38,29 @@ const mapDispatchToProps = (dispatch: Dispatch<Action>): ConnectedDispatch => {
...
```
fragosti, line 61: Longer term you might want this code in the actual component just because we'll need to be awaiting in there to show the loading animations.

ts/redux/async_dispatcher.ts
```diff
@@ -58,15 +59,47 @@ export const asyncDispatcher = {
...
+        providerState: ProviderState,
+        networkId: Network,
+        amountToStakeBaseUnits: BigNumber,
+        dispatcher: Dispatcher,
+    ) => {
+        const { provider } = providerState;
+        const ownerAddress = (providerState.account as AccountReady).address;
+        const gasInfo = await backendClient.getGasInfoAsync();
+        const gasPriceInGwei = new BigNumber(gasInfo.fast / 10);
+        const gasPriceInWei = gasPriceInGwei.multipliedBy(1000000000);
```
fragosti: I know this is in `blockchain.ts` but seems useful.. maybe it should be another method on backendClient. Also just for reference here is similar code in instant https://github.com/0xProject/0x-monorepo/blob/3973dec9944a10c62ec6c3d15d973a00a6ea45a6/packages/instant/src/util/gas_price_estimator.ts#L31
The `estimatedTimeMs` should be useful for the countdown.

ts/redux/async_dispatcher.ts
```diff
@@ -58,15 +59,47 @@ export const asyncDispatcher = {
...
+
+        const contractAddresses = getContractAddressesForNetworkOrThrow(networkId as number);
+        const erc20ProxyAddress = contractAddresses.erc20Proxy;
+        const zrxTokenContract = new ERC20TokenContract(contractAddresses.zrxToken, provider);
+
+        const currentAllowance = await zrxTokenContract.allowance.callAsync(ownerAddress, erc20ProxyAddress);
+
+        // TODO: some information modal needed?
+        if (currentAllowance.isLessThan(amountToStakeBaseUnits)) {
+            // tslint:disable:await-promise
```
fragosti, line 97: Don't think you need this comment.

ts/redux/async_dispatcher.ts
```diff
@@ -58,15 +59,47 @@ export const asyncDispatcher = {
...
+            await zrxTokenContract.approve.awaitTransactionSuccessAsync(erc20ProxyAddress, amountToStakeBaseUnits, {
+                from: ownerAddress,
+                gasPrice: gasPriceInWei,
+            });
```
fragosti: We usually do unlimited allowance, so it is more of a binary thing. 
https://github.com/0xProject/0x-starter-project/blob/f426e1c60e5bceb8848ca7fb618075289df9be09/src/scenarios/fill_order_sra.ts#L59

fragosti (commented):
Are you sure we need to set our allowances before we can stake? What does the flow look like?

ts/redux/async_dispatcher.ts
kimpers: Good point, I was actually initially looking at moving that one over but figured I should use what is available instead ♻️

ts/redux/async_dispatcher.ts
kimpers: You are right. However without it TSLint shows a warning for `Invalid 'await' of a non-Promise value. (await-promise)` because this method returns a `PromiseWithTransactionHash` object not a normal `Promise`.

ts/containers/staking/wizard/wizard.ts
kimpers: Makes sense, I will iterate on this logic after finishing the upgrade 👍

ts/components/staking/wizard/wizard_flow.tsx
kimpers: Yeah, exactly.

kimpers:
> Are you sure we need to set our allowances before we can stake? What does the flow look like?

Yeah I think so, I can verify though. The flow is: (allowance?) -> deposit ZRX into contract -> move ZRX to a pool.

I'm hoping I can do deposit & move stake in one transaction using the batch function on the staking contracts, but I have not yet verified that.

fragosti (approved):

PR #71, opened 2019-12-03
Implement and use getRecommendedStakingPools

fragosti:
Relates to https://github.com/0xProject/0x-api/pull/28

The UX is still rough but the algo works. Maybe we should figure out how to truncate it to 2 options?

Screen Shot 2019-12-02 at 8 09 53 PM

### DIFF
--- a/ts/types.ts
+++ b/ts/types.ts
@@ -1097,7 +1097,7 @@ export interface StakingAPIPoolsResponse {
     stakingPools: PoolWithStats[];
 }
 
-export interface StakingPoolReccomendation {
+export interface StakingPoolRecomendation {
     pool: PoolWithStats;
     zrxAmount: number;
 }
--- a/ts/types.ts
+++ b/ts/types.ts
@@ -1086,7 +1086,8 @@ export interface EpochPoolStats {
     zrxStaked: number;
     operatorShare: number;
     makerAddresses: string[];
-    protocolFeesGeneratedInEth: number;
+    totalProtocolFeesGeneratedInEth: number;
+    sevenDayProtocolFeesGeneratedInEth: number;
     approximateStakeRatio: number;
 }
 
--- a/ts/types.ts
+++ b/ts/types.ts
@@ -1079,6 +1079,7 @@ export interface Pool {
 export interface PoolWithStats extends Pool {
     currentEpochStats: EpochPoolStats;
     nextEpochStats: EpochPoolStats;
+    sevenDayProtocolFeesGeneratedInEth: number;
 }
 
 export interface EpochPoolStats {
@@ -1087,7 +1088,6 @@ export interface EpochPoolStats {
     operatorShare: number;
     makerAddresses: string[];
     totalProtocolFeesGeneratedInEth: number;
-    sevenDayProtocolFeesGeneratedInEth: number;
     approximateStakeRatio: number;
 }
 
--- a/ts/types.ts
+++ b/ts/types.ts
@@ -1096,3 +1096,8 @@ export interface StakingAPIPoolsResponse {
     approximateNextEpoch: Epoch;
     stakingPools: PoolWithStats[];
 }
+
+export interface StakingPoolReccomendation {
+    pool: PoolWithStats;
+    zrxAmount: number;
+}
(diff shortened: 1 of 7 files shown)


ts/utils/staking_utils.ts
```diff
@@ -0,0 +1,55 @@
...
+    numIterations: number;
+}
+
+export const stakingUtils = {
+    getRecommendedStakingPools: (amountZrxToStake: number, pools: PoolWithStats[], opts?: Partial<GetRecommendedStakingPoolsOptions>): StakingPoolReccomendation[] => {
+        if (!pools || amountZrxToStake === 0) {
+            return [];
+        }
+        const { alpha, numIterations } = {
+            alpha: 2 / 3,
```
alexkroeger: alpha needs to be obtained from the DB or Web3

ts/types.ts
```diff
@@ -1095,3 +1096,8 @@ export interface StakingAPIPoolsResponse {
     approximateNextEpoch: Epoch;
     stakingPools: PoolWithStats[];
 }
+
+export interface StakingPoolReccomendation {
```
kimpers:
```suggestion
export interface StakingPoolRecommendation {
```

ts/utils/staking_utils.ts
```diff
@@ -0,0 +1,55 @@
...
+            sevenDayProtocolFeesGeneratedInEth: pool.currentEpochStats.sevenDayProtocolFeesGeneratedInEth,
+            zrxStaked: pool.nextEpochStats.zrxStaked,
+        }));
+        const stakingDecisions: { [poolId: string]: number } = {};
+        for (let i = 0; i < numIterations; i++) {
+            const totalStake = _.sumBy(poolsSummary, p => p.zrxStaked);
+            const totalProtocolFees = _.sumBy(poolsSummary, p => p.sevenDayProtocolFeesGeneratedInEth);
+            const adjustedStakeRatios: number[] = [];
+            for (const pool of poolsSummary) {
+                const stakeRatio = (pool.zrxStaked / totalStake) / (totalProtocolFees > 0 ? pool.sevenDayProtocolFeesGeneratedInEth / totalProtocolFees : 1);
```
kimpers: I think this may have slightly better readability
```suggestion
                const stakeRatio = (pool.zrxStaked / totalStake) / (pool.sevenDayProtocolFeesGeneratedInEth / (totalProtocolFees  || 1));
```

ts/utils/staking_utils.ts
```diff
@@ -0,0 +1,55 @@
...
+                const adjStakeRatio = Math.pow(((1 - alpha) / (1 - pool.operatorShare)), (1 / alpha)) * stakeRatio;
+                adjustedStakeRatios.push(adjStakeRatio);
+            }
+            const bestPoolIndex = adjustedStakeRatios.indexOf(_.min(adjustedStakeRatios));
+            const bestPool = poolsSummary[bestPoolIndex];
+            stakingDecisions[bestPool.poolId] = stakingDecisions[bestPool.poolId] || 0;
+            stakingDecisions[bestPool.poolId] += amountZrxToStake / numIterations;
+            bestPool.zrxStaked += amountZrxToStake + (amountZrxToStake / numIterations);
```
kimpers: I don't know the algorithm specification so I might be way off here but is this supposed to add the full `amountZrxToStake` on every iteration?

kimpers (changes requested):

chriskalani:
@fragosti that list is supposed to scroll. Should look like this:

General Entrypoint 3 (multipool)

ts/utils/staking_utils.ts
fragosti: @alexkroeger

fragosti: Nice catch!

ts/utils/staking_utils.ts
fragosti: Let me see if that is possible. Otherwise I don't expect this value to change very often correct?

kimpers (approved):

johnrjj (approved):

ts/utils/staking_utils.ts
fragosti: I spoke to greg. I don't think we'll be changing this very often so not worth fetching. Can always update the frontend.

PR #100, opened 2019-12-16
Staking Wizard Recommended Pools Calc w/ BigNumber

johnrjj:
Previously we were overflowing our floating point (e.g. 400 / 3 ended up producing `133.33333333333335`ish and ended up with a total of `400.00000000000004` (which if your balance is 400 is going to cause the transaction to fail)

Note: I had to use decimal.js (a more feature complete version of BigNumber) for this as BigNumber does not (and will not) support decimal exponentials. I figure that's okay for this as it's a completely isolated module we can swap out as needed. I won't let the decimal.js types leak from the module.

### DIFF
--- a/ts/utils/staking_utils.ts
+++ b/ts/utils/staking_utils.ts
@@ -1,10 +1,11 @@
-// import { BigNumber } from '@0x/utils';
+// Need to use decimal.js (more complete version of BigNumber) as bignumber doesn't allow for decimal exponents
+import { Decimal as BigNumber } from 'decimal.js';
 import * as _ from 'lodash';
-import {Decimal as BigNumber } from 'decimal.js'; // need to use decimal.js as bignumber doesn't 
-import { PoolWithStats, StakingPoolRecomendation } from '../types';
 
-import { constants } from './constants';
+import { PoolWithStats, StakingPoolRecomendation } from 'ts/types';
+
 import { configs } from 'ts/utils/configs';
+import { constants } from 'ts/utils/constants';
 
 interface PoolStatSummary {
     poolId: string;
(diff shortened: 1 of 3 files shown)


johnrjj:
@kimpers -- if you get some free time tomorrow, can you spot check this? it should be equivalent to the previous implementation, but more accurate. (interestingly enough, I'm getting different results, even though the math is the same as far as I can tell -- I'm fairly certain the native js math impl was producing incorrect results)

ts/utils/staking_utils.ts
```diff
@@ -32,26 +35,51 @@ export const stakingUtils = {
...
-            stakingDecisions[bestPool.poolId] += amountZrxToStake / numIterations;
+            stakingDecisions[bestPool.poolId] = stakingDecisions[bestPool.poolId] || new BigNumber(0);
+            stakingDecisions[bestPool.poolId] = stakingDecisions[bestPool.poolId].plus((new BigNumber(amountZrxToStake).div(numIterations)));
             bestPool.zrxStaked += (amountZrxToStake / numIterations);
         }
+
         const recs = Object.keys(stakingDecisions).map(poolId => ({
             pool: pools.find(pool => pool.poolId === poolId),
-            zrxAmount: stakingDecisions[poolId],
+            zrxAmount: Number(stakingDecisions[poolId].toFixed(configs.AMOUNT_DISPLAY_PRECSION)),
```
kimpers, line 66: Correct me if I'm wrong but I think `toFixed` may round the value which could cause problems with the total not adding up. Maybe it makes sense to convert it to our BigNumber and return that?

kimpers (approved):

johnrjj:
Not needed right now.

PR #120, opened 2020-01-02
Update to instant v3 url

dekz:
Updates the URLs pointing to instant to serve the v3 version.


johnrjj:
@dekz Is this something we should land before the staking dashboard goes live?

johnrjj:
This was included in https://github.com/0xProject/website/pull/162 which is now in `development`

PR #122, opened 2020-01-03
Fix staking portal linting issues

kimpers:
I got tired of having to use `--no-verify` so I fixed our linting issues, I also ran prettier on every file I had to edit manually.


johnrjj (approved):
🙌

PR #130, opened 2020-01-04
Fix horizontal overflow in safari by disabling custom scroll styles in webkit

johnrjj:
Prior PR (https://github.com/0xProject/website/pull/129) was good, did not fix the horizontal overflow issue in Safari desktop.

Turns out this is caused because of the custom styling of the scrollbar in webkit. Behavior in Safari is a bit fragile around this; Instead of messing around with calculating widths correctly, let's just disable this for now (at least until post-launch). 

This fixes the website in Safari (noticeable with the trackpad).

### DIFF
--- a/less/all.less
+++ b/less/all.less
@@ -1,4 +1,3 @@
-// Safari can slightly overflow the page width without this
 html, body {
     max-width: 100%;
 }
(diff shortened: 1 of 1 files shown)


kimpers (approved):

kimpers:
Sorry for being slow on reviewing this, I thought this was the previous pr that I had already reviewed 🤦‍♂

johnrjj:
No worries !

PR #134, opened 2020-01-06
Adding staking lottie files for animations

johnrjj:
Not wired up yet, just wanted to add the static assets. Will follow up with another PR when I get a chance.

Also moves the animation_loader into a central location (rather than inside of `/docs`) as I found it to be reusable.


kimpers (approved):
LGTM 👍

PR #142, opened 2020-01-08
Adds useStakingWizard hook, refactors wizard to be router driven

johnrjj:
This PR implements `useStakingWizard` for guiding the majority of the wizard. It is now primarily route driven  (with each primary step holding some internal state). Most things remain largely declarative (what you see is based on the route as the source of truth). This also allows us to manually move between steps (`goToNextStep`) , or do it in a declarative way. Also this gives us a back button that will naturally work with browsers (will implement in next PR) as well as sets us up nicely for animations (upcoming today hopefully!)

I'm happy to own this part while we release, should something go wrong with it -- I've kicked the tires on it for a bit and I think its fine from my own testing. Or we can revert it without much hassle if we need to.

Here's everything still working -- but note the router changing the wizard `step` query param.
demo

### DIFF
--- a/ts/components/staking/wizard/wizard_flow.tsx
+++ b/ts/components/staking/wizard/wizard_flow.tsx
@@ -258,6 +258,7 @@ export interface StakingInputPaneProps {
     stakingPools: any;
     zrxBalance: BigNumber;
     onOpenConnectWalletDialog: () => any;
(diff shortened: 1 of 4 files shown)


kimpers (approved):
Awesome! 🍻

PR #143, opened 2020-01-08
Refactor web3 wallet connect to hook and keep account state updated on changes

kimpers:


ts/hooks/use_wallet.ts
```diff
@@ -0,0 +1,105 @@
...
+        const networkId = await providerState.web3Wrapper.getNetworkIdAsync();
+        if (networkId !== currentNetworkId) {
+            dispatcher.updateNetworkId(networkId);
+        }
+
+        await asyncDispatcher.fetchAccountInfoAndDispatchToStoreAsync(providerState, dispatcher, networkId, true);
+
+        const newProvider = providerState.provider as any;
+        if (newProvider.on) {
+            accountsChangedRef.current = handleAccountsChange;
```
kimpers, line 78: This is kinda funky but I couldn't think of a better way to do this in order to preserve the fn reference to be able to unsubscribe in the future. I'm open to suggestions.

johnrjj: That's totally fine -- I use useRef as a box like that too (that's pretty standard to hold on to things that you don't care about rendering updates for [that's really the only footgun - useRef assignments don't trigger a rerender])

ts/hooks/use_wallet.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    const logoutWallet = useCallback(() => {
+        cleanupCurrentProvider();
+        dispatcher.setAccountStateLoading();
+    }, [cleanupCurrentProvider, dispatcher]);
+
+    return {
+        connectToWallet: useCallback(() => {
+            connectToWallet().catch((err: Error) => {
+                logUtils.warn(`Failed to connect wallet ${err}`);
```
johnrjj, line 100: Totally unrelated (but reminded me nonetheless) -- do we have a frontend error monitoring service? Sentry/BugSnag? I would like to hook a few of these critical errors up to some sort of reporting mechanism.

johnrjj (approved):
Nice! Really liking the hooks refactor. 

Thanks for using useCallback everywhere, as it prevents a lot of unnecessary renders. 😄

ts/hooks/use_wallet.ts
kimpers: Not to my knowledge unfortunately 😢  Would love to get something like Sentry in, have been thinking about it myself but wasn't able to anything about it yet.

PR #149, opened 2020-01-08
Adds newsletter subscription functionality to staking success pane

johnrjj:
I had to fix the Input a little bit

+ Prettier

### DIFF
--- a/ts/pages/staking/wizard/newsletter.tsx
+++ b/ts/pages/staking/wizard/newsletter.tsx
@@ -93,7 +93,7 @@ export const Newsletter = () => {
                 return;
             }
             try {
-                await backendClient.subscribeToNewsletterAsync(email);
+                await backendClient.subscribeToNewsletterAsync(email.trim());
             } catch (e) {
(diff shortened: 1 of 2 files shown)


ts/pages/staking/wizard/newsletter.tsx
```diff
@@ -82,19 +85,26 @@ const StyledButton = styled(Button)`
...
-    };
+    const [email, setEmail] = React.useState<string>('');
+    const onSubmit = React.useCallback(
+        async (event: React.FormEvent<HTMLFormElement>) => {
+            event.preventDefault();
+            if (!email) {
+                return;
+            }
+            try {
+                await backendClient.subscribeToNewsletterAsync(email);
```
kimpers: NIT: might be worth trimming the email before sending it off to the server, I don't think we do any validation in the backend client
```suggestion
                await backendClient.subscribeToNewsletterAsync(email.trim());
```

kimpers (approved):

PR #153, opened 2020-01-08
Fixes active pool filter logic to use OR instead of AND

johnrjj:


johnrjj:
@kimpers Going to merge this real quick as it's a super quick change and they would like to see it on dogfood.

PR #172, opened 2020-01-09
Removes token approval info pane (left side)

johnrjj:
Too noisy, let's remove and just use the previous step info pane.


kimpers (approved):
LGTM 👍

PR #175, opened 2020-01-09
Use next epoch for wizard flow shared and staked params

kimpers:


johnrjj (approved):

johnrjj:
Oh did I miss these? I can deploy this real quick

PR #180, opened 2020-01-09
Update formatNumber postfix to use 2 digits on M postfix symbol

johnrjj:
yay for tests


johnrjj:
ok last ninja merge :p

PR #182, opened 2020-01-09
Adds cumulative sum to graph

johnrjj:
By request. This is probably a short to medium term solution, eventually we'll need pagination on epochs and this won't work client side.


kimpers (approved):
LGTM

johnrjj:
Not going to implement this right now. Closing.

PR #187, opened 2020-01-10
Grab current epoch rewards from api instead of with web3

kimpers:
This is kinda 🤦‍♂ We had the correct value available in the api under a different name, so there is no point in fetching this from the smart contract.


johnrjj (approved):
Love a red diff :shipit: 

I can merge/release, it's the weekend your time :)

PR #201, opened 2020-01-15
Use Bamboo Relay for CFL metrics

kimpers:
image

Since there is a problem with the Radar metrics use Bamboo instead for now.


johnrjj:
I was like 'is that eth/dai right?' 🔥

johnrjj (approved):

PR #204, opened 2020-01-16
Increase granularity on time to epoch end on staking home

kimpers:
image

Quick fix per internal request, would be cool to make this interactively count down in the future but I think this is better than what we have now.


johnrjj (approved):

johnrjj:
Merging to fix

PR #205, opened 2020-01-17
Reverts fix/use-api-data-for-rewards (#187)

johnrjj:


ts/utils/configs.ts
```diff
@@ -7,11 +7,11 @@ export const configs = {
     AMOUNT_DISPLAY_PRECSION: 5,
     BACKEND_BASE_PROD_URL: 'https://website.api.0x.org',
     BACKEND_BASE_STAGING_URL: 'https://staging-website-api.0x.org',
-    BACKEND_BASE_DEV_URL: 'https://localhost:3001',
+    BACKEND_BASE_DEV_URL: 'https://website.api.0x.org',
     API_BASE_PROD_URL: 'https://api.0x.org',
     API_BASE_STAGING_URL: 'https://staging.api.0x.org',
     API_BASE_KOVAN_URL: 'https://kovan.api.0x.org',
-    API_BASE_DEV_URL: 'http://localhost:3000',
+    API_BASE_DEV_URL: 'https://api.0x.org',
```
kimpers: These should not be here, I will clean up and merge so `development` is not dirty

kimpers (approved):
Double 🤦‍♂ on this one 😄 

We should look into fixing the value in api, it was so nice having everything load instantly rather than this number popping up later.

ts/utils/configs.ts
johnrjj: thanks good catch! was testing late last night haha

PR #208, opened 2020-01-19
Fix lowercase ETH address before sending it to API

kimpers:
Workaround until 0xProject/0x-api#76 is merged and deployed.

Will merge this to get it out and follow for any comments.


johnrjj:
LGTM thanks for shipping this :)

PR #211, opened 2020-01-21
Staking home pools sorting

kimpers:
**Desktop**
image

**Mobile**
image
Not sure if we want to show it on mobile, it's a bit awkward but I think it still provides value. Wdyt?

**NOTE:** view with whitespace changes disabled to avoid giant prettier generated diffs 😄


johnrjj (approved):
LGTM, I think it’s fine to keep on mobile. Great job!

PR #222, opened 2020-01-28
Hotfix/doc guide links

johnrjj:
Will actually fix this after launch. Not sure how docs actually work so just patching for now.

Background: the ABCS of contract fillable liquidity does not have an external url or regular url when it gets to the Resource file. It seems to be the only one behaving this way. Not sure how this happened (randomly started failing). The entire docs site crashes if a <Link> is missing the `to` prop, so as a hack we interpret the guide url from the metadata.


johnrjj:
Last ninja, I promise

PR #227, opened 2020-01-30
Move tracking staking events into Google Analytics

kimpers:


ts/utils/googleAnalytics.ts
```diff
@@ -0,0 +1,3 @@
+export const trackEvent = (eventName: string, params?: object) => {
+    (window as any).gtag('event', eventName, params);
+};
```
kimpers: Might be a bit silly to put this in its own file but I think that we will want to track more things if we continue with GA exclusively.

In the future we can hopefully remove the entire Heap integration and the `analytics` module but I'm leaving it in for now until we are fully settled on GA.

johnrjj: Sounds good! (super tiny nit: should this file be `google_analytics.ts` to follow our naming scheme? i don't mind either way)

johnrjj (approved):

ts/utils/googleAnalytics.ts
kimpers: Whoops, too many years of `camelCase` 🤦‍♂

PR #237, opened 2020-02-13
Confirmation dialog for removing stake

kimpers:
**Desktop**
image

**Mobile**
image

### DIFF
--- a/ts/components/staking/remove_stake_dialog.tsx
+++ b/ts/components/staking/remove_stake_dialog.tsx
@@ -1,0 +1,163 @@
+import { BigNumber } from '@0x/utils';
+import { DialogContent, DialogOverlay } from '@reach/dialog';
+import '@reach/dialog/styles.css';
+import React, { FC } from 'react';
+import styled from 'styled-components';
+
+import { Button } from 'ts/components/button';
+import { Icon } from 'ts/components/icon';
+import { Heading, Paragraph } from 'ts/components/text';
+
+import { zIndex } from 'ts/style/z_index';
+import { PoolWithStats } from 'ts/types';
+import { colors } from 'ts/utils/colors';
+import { formatEther, formatZrx } from 'ts/utils/format_number';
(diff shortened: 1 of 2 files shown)


ts/pages/account/dashboard.tsx
```diff
@@ -83,9 +84,11 @@ interface PendingAction {
     amount: number;
 }
 
-// TODO(johnrjj) - After going to a pool that has been staked in, hitting the 'Remove' button,
-// and running the transaction, the balances do not update, even after a good amount of time (30min+ etc)
-// Is there a way we can query web3 directly?? (tl;dr 'Pending' removes are flaky/not sure if they are working)
```
kimpers, line 88: @johnrjj can you confirm that this is not an issue anymore? It works fine for me but I'm not sure I fully understand what the problem was.

ts/components/staking/remove_stake_dialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+        @media (max-width: 768px) {
+            min-height: 100vh;
+            width: 100vw;
+            margin: 0;
+            padding: 30px;
+
+            border: none;
+        }
+    }
+`;
```
kimpers, line 163: Mostly copy-pasted from `change_pool_dialog`, it is small enough that it didn't feel worth it creating a generic dialog component but if we start adding more of these we should.

ts/pages/account/dashboard.tsx
johnrjj: This must've been outdated -- everything looks good now! :)

ts/components/staking/remove_stake_dialog.tsx
johnrjj: that's my preferred level of abstraction too 😄

johnrjj (approved):
LGTM -- 0x 4 lyfe!

PR #239, opened 2020-02-13
Margin trading guide with 0x API

dave4506:
^ 

If you have the time please skim the codebase https://github.com/0xProject/0x-api-starter-guide-code for any typos or oversights on my part.

### DIFF
--- a/mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
+++ b/mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
@@ -70,14 +70,14 @@ constructor (
         )
         public
     {
-        EXCHANGE = IExchange(_exchange);
-        COMPTROLLER = IComptroller(_comptroller);
-        CDAI = ICERC20(_ceth);
-        CETH = ICETH(_cdai);
-        WETH = IEtherToken(_weth);
-        DAI = IERC20Token(_dai);
+        exchange = IExchange(_exchange);
+        comptroller = IComptroller(_comptroller);
+        cdai = ICERC20(_ceth);
+        ceth = ICETH(_cdai);
+        weth = IEtherToken(_weth);
+        dai = IERC20Token(_dai);
 
-        OWNER = msg.sender;
+        owner = msg.sender;
 
         // Enter markets
         _enterMarkets();
@@ -88,9 +88,9 @@ constructor (
 Let's break down a few of the references we are going to be interacting with:
 
 
-1. `EXCHANGE`: the main 0x protocol smart contract to interact with
-2. `COMPTROLLER`: the main compound finance smart contract to interact with
-3. `CDAI`, `CETH`: 'c' tokens used by compound finance to represent assets in the protocol and accrued interests
+1. `exchange`: the main 0x protocol smart contract to interact with
+2. `comptroller`: the main compound finance smart contract to interact with
+3. `cdai`, `ceth`: 'c' tokens used by compound finance to represent assets in the protocol and accrued interests
 
 `_enterMarkets()` is a function that allows the contract to interact with Compound Finance for the ETH and DAI money markets.
 
@@ -110,7 +110,7 @@ function open(ZeroExQuote memory quote)
 }
 ```
 
-The function will only allow the `OWNER` to open a position and only when the smart contract has no currently opened positions. The parameter `quote` will contain the 0x API response to marketBuy WETH with DAI.
+The function will only allow the `owner` to open a position and only when the smart contract has no currently opened positions. The parameter `quote` will contain the 0x API response to marketBuy WETH with DAI.
 
 #### Determining amount of ETH to deposit into Compound finance
 
@@ -139,7 +139,7 @@ With `positionBalance`, we can then deposit and mint cETH with Compound Finance.
 // 2. increase position by msg.value - protocolFeeAmount
 positionBalance = msg.value.safeSub(quote.protocolFeeAmount);
 // 3. mint collateral in compound
-CETH.mint.value(positionBalance)()
+ceth.mint.value(positionBalance)()
 ...
 ```
 
@@ -148,11 +148,11 @@ Now that we have `accountLiquidity` (borrowing power in Compound Finance), let's
 ```
 ...
 // 4. borrow token
-assert(CDAI.borrow(quote.sellAmount) == 0);
+require(cdai.borrow(quote.sellAmount) == 0, "Borrow DAI failed.");
 ...
 ```
 
-Compound finance returns a non-zero value if an error occurs during the function call, with an `assert` we can proceed with the function only when the function doesn't throw any errors.
+Compound finance returns a non-zero value if an error occurs during the function call, with a `require` we can proceed with the function only when the function doesn't throw any errors.
 
 #### Swaping for ETH with 0x liquidity 
 
@@ -161,7 +161,7 @@ To swap with 0x, the 0x exchange contracts will need to have access to move DAI 
 ```solidity
 ...
 // 4. approve 0x exchange to move DAI
-_approve(address(DAI), _getZeroExApprovalAddress());
+_approve(address(dai), _getZeroExApprovalAddress());
 ...
 ```
 
@@ -170,20 +170,20 @@ Since the quote is passed off-chain, it is worth verifying that the assets you i
 ```solidity
 ...
 // 5. verify quote is valid
-require(quote.sellToken == address(WETH), "not buying WETH");
-require(quote.buyToken == address(DAI), "not buying WETH");
+require(quote.sellToken == address(weth), "not buying WETH");
+require(quote.buyToken == address(dai), "not buying WETH");
 ...
 ```
 
-Now that we have verified and set the approvals to perform a swap with 0x protocol, let’s call `EXCHANGE` with `callDataHex` provided in the quote.
+Now that we have verified and set the approvals to perform a swap with 0x protocol, let’s call `exchange` with `callDataHex` provided in the quote.
 
 ```solidity
 ...
 // 6. execute swap
-(bool success, bytes memory data) = address(EXCHANGE).call.value(quote.protocolFeeAmount)(quote.calldataHex);
+(bool success, bytes memory data) = address(exchange).call.value(quote.protocolFeeAmount)(quote.calldataHex);
 require(success, "Swap not filled.");
 // 7. decode fill results
-LibFillResults.FillResults memory fillResults = abi.decode(data (LibFillResults.FillResults));
+LibFillResults.FillResults memory fillResults = abi.decode(data, (LibFillResults.FillResults));
 ...
 ```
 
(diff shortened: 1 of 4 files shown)


fragosti (approved):

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+        address _exchange,
+        address _comptroller,
+        address _cdai,
+        address _dai,
+        address _ceth,
+        address _weth
+        )
+        public
+    {
+        EXCHANGE = IExchange(_exchange);
```
jalextowle: These should be lower-case because they aren't actually constants (we're setting them right here). 

In our production smart contracts, we would actually prefer to use constants for all of these addresses. If you check out the staking contracts, you'll find the deployment constants pattern. That is how we would actually write this, so you'll need to make a call between the added complexity of the pattern vs. the fact that the current style of initialization is sub-optimal for production smart contracts.

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+function open(ZeroExQuote memory quote)
+  public
+  payable
+  onlyOwner
+  onlyWhenClosed
+  returns (uint256 positionBalance, uint256 wethBalance, uint256 borrowBalance)
+{
+    ...
+  // 1. increase position by msg.value - protocolFeeAmount
+  positionBalance = msg.value.safeSub(quote.protocolFeeAmount);
```
jalextowle: The `protocolFeeAmount` can be calculated as `protocolFeeAmount = tx.gasPrice.safeMul(exchange.protocolFeeMultiplier());`. As I mentioned offline, I think this should be calculated on-chain for DevEx reasons.

jalextowle: Thinking about this more, this guide might be the right place to introduce the various ways in which protocol fees can be handled by a calling contract. Maybe we should take WETH from `msg.sender` if `msg.value < protocolFeeAmount` instead of reverting.

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+CETH.mint.value(positionBalance)()
+...
+```
+
+Now that we have `accountLiquidity` (borrowing power in Compound Finance), let's borrow DAI from Compound Finance. We will be borrowing the exact amount of DAI we intend to sell with 0x liquidity. Off-chain, `sellAmount` is calculated based on our desired leverage.
+
+```
+...
+// 4. borrow token
+assert(CDAI.borrow(quote.sellAmount) == 0);
```
jalextowle: We should use `require` instead of `assert` in almost all cases. We should only use `assert` for things that shouldn't ever happen in production (and even then, there is still an argument to use `require`). Additionally, we should add a revert reason. You can do this by adding a string argument to `require`:

```
require(true, "This is the error message");
```

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+...
+```
+
+Since the quote is passed off-chain, it is worth verifying that the assets you intend to sell and buy.
+
+```solidity
+...
+// 5. verify quote is valid
+require(quote.sellToken == address(WETH), "not buying WETH");
+require(quote.buyToken == address(DAI), "not buying WETH");
```
jalextowle: Is this the right error message?

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+
+Now that we have verified and set the approvals to perform a swap with 0x protocol, let’s call `EXCHANGE` with `callDataHex` provided in the quote.
+
+```solidity
+...
+// 6. execute swap
+(bool success, bytes memory data) = address(EXCHANGE).call.value(quote.protocolFeeAmount)(quote.calldataHex);
+require(success, "Swap not filled.");
+// 7. decode fill results
+LibFillResults.FillResults memory fillResults = abi.decode(data (LibFillResults.FillResults));
```
jalextowle: Will this compile? Are you missing a comma?

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+```
+
+#### Updating the contract state
+
+With that, we have just swapped DAI for WETH, effectively opening our margin position. Before we finish this function, let’s update the contract state and return some results to the user. From the 0x contract call, we have a receipt that contains the actual amount of buy and sell token swapped.
+
+```solidity
+...
+// 8. position size increase by bought amount of WETH (makerAssetFilledAmount)
+positionBalance += fillResults.makerAssetFilledAmount;
```
jalextowle: We should never use arithmetic operators. Let's use `positionBalance = positionBalance.safeAdd(fillResults.makerAssetFilledAmount)` here.

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+        // 5. verify quote is valid
+        require(quote.sellToken == address(WETH), "not buying WETH");
+        require(quote.buyToken == address(DAI), "not buying WETH");
+        // 6. execute swap
+        (bool success, bytes memory data) = address(EXCHANGE).call.value(quote.protocolFee)(quote.calldataHex);
+        require(success, "Swap not filled.");
+        // 7. decode fill results
+        LibFillResults.FillResults memory fillResults = abi.decode(data, (LibFillResults.FillResults));
+        // 8. position size increase by bought amount of WETH
+        positionBalance += fillResults.makerAssetFilledAmount;
```
jalextowle: Ditto here

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+
+Similar to swapping from DAI to WETH in `open`, we will need to do a few things before swapping:
+
+```solidity
+...
+// 1. approve for swap
+_approve(address(WETH), _getZeroExApprovalAddress());
+// 2. verify swap
+uint256 wethBalance = WETH.balanceOf(address(this));
+uint256 daiBorrowBalance = CDAI.borrowBalanceCurrent(address(this)); // TODO
```
jalextowle: Why do we have a `TODO` here?

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+    require(daiBorrowBalance < quote.buyAmount, "not enough DAI to repay");
+    // execute swap
+    (bool success, bytes memory data) = address(EXCHANGE).call.value(quote.protocolFee)(quote.calldataHex);
+    require(success, "Swap not filled.");
+    // decode results
+    LibFillResults.FillResults memory fillResults = abi.decode(data, (LibFillResults.FillResults));
+    // return back DAI
+    _approve(address(DAI), address(CDAI));
+    // approve Compound Finance CDAI
+    assert(CDAI.repayBorrow(fillResults.makerAssetFilledAmount) == 0);
```
jalextowle: As stated above, let's use `require` (unless I'm missing context), and add an error message.

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+    // get back ETH
+    assert(CETH.redeem(CETH.balanceOfUnderlying(address(this))) == 0); // TODO correct?
```
jalextowle: Ditto + remove the `TODO` please

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+    onlyWhenOpen
+    returns (uint ethBalance)
+{
+    // approve for swap
+    _approve(address(WETH), _getZeroExApprovalAddress());
+    // verify swap
+    uint256 wethBalance = WETH.balanceOf(address(this));
+    uint256 daiBorrowBalance = CDAI.borrowBalanceCurrent(address(this)); // TODO
+
+    require(wethBalance > quote.sellAmount, "not enough to swap");
```
jalextowle: We would typically bundle up a bunch of these `require` statements into a function. That may make the guide harder to read, but it's something to consider.

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -0,0 +1,573 @@
...
+    public
+    onlyOwner
+    onlyWhenOpen
+    returns (uint ethBalance)
+{
+    // approve for swap
+    _approve(address(WETH), _getZeroExApprovalAddress());
+    // verify swap
+    uint256 wethBalance = WETH.balanceOf(address(this));
+    uint256 daiBorrowBalance = CDAI.borrowBalanceCurrent(address(this)); // TODO
```
jalextowle: nit: `TODO`

jalextowle (changes requested):

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
dave4506: Good point, I was considering this guide to gloss over those subjects, namely this is the first guide for people to piece together how 0x API can fulfill roles that of Uniswap. I feel like all the nuances with protocol fees can be better handled in a dedicated guide that contains all that content.

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
dave4506: Making the call to use the pattern above to make it easier to gloss over and understand what is happening. Mainly, if people want to develop on a testnet, I'm hoping readers do not have to manipulate the sol file and migration/deployment system to do so. Seem like this pattern creates that better dev ex (for readability). Will lower case them per your comment.

fragosti (approved):

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
jalextowle: Hmm, yeah that's probably a good call. I think having that guide would be really useful though.

PR #243, opened 2020-02-18
WalletLink interim integration

kimpers:
Super hacky but works for now


ts/utils/providers/provider_factory.ts
```diff
@@ -1,12 +1,21 @@
...
 import { ZeroExProvider } from 'ethereum-types';
+import _ from 'lodash';
+import WalletLink from 'walletlink';
 
 import { Maybe, Network } from 'ts/types';
 import { configs } from 'ts/utils/configs';
 
 const { PUBLIC_NODE_URLS_BY_NETWORK_ID } = configs;
 
+const LOGO_URL_0x = 'https://0x.org/images/0x_logo.png';
```
kimpers, line 12: This is shown inside the Coinbase Wallet app

ts/hooks/use_wallet.ts
```diff
@@ -63,45 +64,59 @@ export const useWallet = () => {
...
+            const providerState: ProviderState =
+                providerOverride === Providers.WalletLink
+                    ? providerStateFactory.getInitialProviderStateFromWalletLink(currentNetworkId)
+                    : providerStateFactory.getInitialProviderState(currentNetworkId);
 
-        const networkId = await providerState.web3Wrapper.getNetworkIdAsync();
-        if (networkId !== currentNetworkId) {
-            dispatcher.updateNetworkId(networkId);
-        }
+            dispatcher.updateProviderState(providerState);
```
kimpers, line 75: This little bug was pretty tricky to hunt down :feelsgood: Previously we would set the `providerState` when initializing the redux store and then never update it. That meant that it was either set to use MetaMask or a read-only RPC provider in the case where there was no MM.

When connecting to another wallet like Wallet Link it would successfully initialize the new provider and providerState but never update the value in the store so when you later try sending a transaction it's using the wrong provider.

johnrjj: oh gosh

johnrjj (approved):
awesome!! about what I expected code-wise

LGTM; I think this plenty adequate for now -- great job!

PR #245, opened 2020-02-24
Hide vote button on ZEIP if voting has ended

johnrjj:
This whole section of the site uses moment, so I just followed that pattern.

### DIFF
--- a/ts/pages/governance/governance.tsx
+++ b/ts/pages/governance/governance.tsx
@@ -1,5 +1,6 @@
 import { BigNumber } from '@0x/utils';
 import * as _ from 'lodash';
+import moment from 'moment';
 import * as React from 'react';
 import { RouteComponentProps } from 'react-router-dom';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


ts/pages/governance/governance.tsx
```diff
@@ -65,6 +66,11 @@ export class Governance extends React.Component<RouteComponentProps<any>> {
     }
     public render(): React.ReactNode {
         const { isVoteReceived, tally } = this.state;
+
+        const now = moment();
+        const pstOffset = '-0800';
+        const deadlineToVote = this._proposalData?.voteEndDate?.utcOffset(pstOffset);
+        const hasVoteEnded = deadlineToVote?.isBefore(now) || false;
```
johnrjj, line 73: none of this data should ever be undefined (but since it does array index lookup to get the vote, technically can be undefined) -- what's the best boolean default here? I leaned toward the vote being expired if we can't find it/issue loading it

johnrjj: (we also don't enforce `undefined` strict errors in this TS project, so those can be sneaky)

kimpers: Agreed, better to default to vote not active in case of issues.

kimpers (approved):

PR #249, opened 2020-02-28
Allow instant purchases with WalletLink & update balance after buy

kimpers:
Noticed that instant didn't work with Wallet Link after deploying it, also we should try to update the balance after a purchase 😄


johnrjj (approved):
THANKYOU! this is great!

PR #253, opened 2020-03-04
Add bitpie wallet and update Walletlink

kimpers:


johnrjj (approved):

chriskalani:
What is Bitpie?

kimpers:
> What is Bitpie?

It's a wallet based on Trust Wallet that is popular in asia (https://bitpie.com). The problem is that we currently identify it as Trust Wallet which leads to some confusion.

PR #261, opened 2020-03-11
Fixes rounding issue on the account staking page for pool's rewards shared

johnrjj:
This is more of an edge case, as a user doesn't really use the account page to verify the rewards shared (the staking pool page is always accurate), but wanted to fix it. Anyways, we were rounding down at the edge and we shouldn't, this fixes that.


kimpers (approved):
LGTM

PR #269, opened 2020-03-25
Fix staking pool sorting and Dust pool icon size

kimpers:


ts/pages/staking/home.tsx
```diff
@@ -157,6 +157,7 @@ export const StakingIndex: React.FC<StakingIndexProps> = () => {
                                 poolId={pool.poolId}
                                 name={stakingUtils.getPoolDisplayName(pool)}
                                 thumbnailUrl={pool.metaData.logoUrl}
+                                isFullSizeTumbnail={pool.poolId === '12'}
```
kimpers, line 160: This is definitely a hack but I think it's fine for now. The other option would be to change the JSON structure to allow a parameter for controlling this to be passed by the pool operators. But since only 1 has requested this I think we can use this for now.

johnrjj (approved):
LGTM, thank you!

PR #278, opened 2020-04-13
Update config.ts to use alchemy's v2

opaolini:
About
===

This PR updates our provider configuration to the v2 endpoint as Alchemy is migrating their primary nodes to Geth [0],

[0] - https://medium.com/alchemy-api/geth-it-right-c3c27cec43c


kimpers (approved):
Thanks for this! 👍 I tested it locally on Kovan and everything seems to be working as intended.

PR #281, opened 2020-04-16
Fix analytics wallet and staking reporting

kimpers:
Using custom attributes in Google Analytics is actually far more complicated than I initially imagined. Since our flows are rather simple let's just use the `event_label` as a custom value field for now.


johnrjj (approved):
lol, thought this was a matcha PR for a good minute and was fairly confused

LGTM

PR #285, opened 2020-04-21
Feature/add account activity

alexkroeger:
Added live data to the account activity page, and a link to access this page from the account page.

image

image

### DIFF
--- a/ts/pages/account/account_detail.tsx
+++ b/ts/pages/account/account_detail.tsx
@@ -34,7 +34,7 @@ export const AccountDetail: React.StatelessComponent<AccountDetailProps> = ({
                         isWithArrow={true}
                         isAccentColor={true}
                         shouldUseAnchorTag={true}
-                        target="_blank"
+                        target="_self"
                     >
                         Show all activity
                     </Button>
--- a/ts/pages/account/account_detail.tsx
+++ b/ts/pages/account/account_detail.tsx
@@ -5,6 +5,7 @@ import { Button } from 'ts/components/button';
 import { generateUniqueId, Jazzicon } from 'ts/components/ui/jazzicon';
 
 import { colors } from 'ts/style/colors';
+import { WebsitePaths } from 'ts/types';
 import { utils } from 'ts/utils/utils';
 
 export interface AccountDetailProps {
@@ -30,11 +31,10 @@ export const AccountDetail: React.StatelessComponent<AccountDetailProps> = ({
                 </Avatar>
                 <ButtonWrapper>
                     <Button
-                        href="/zrx/account/activity"
+                        to={WebsitePaths.AccountActivity}
                         isWithArrow={true}
                         isAccentColor={true}
                         shouldUseAnchorTag={true}
-                        target="_self"
                     >
                         Show all activity
                     </Button>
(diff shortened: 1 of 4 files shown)


ts/pages/account/account_detail.tsx
```diff
@@ -15,17 +16,30 @@ export const AccountDetail: React.StatelessComponent<AccountDetailProps> = ({
...
+                <ButtonWrapper>
+                    <Button
+                        href="/zrx/account/activity"
+                        isWithArrow={true}
+                        isAccentColor={true}
+                        shouldUseAnchorTag={true}
+                        target="_blank"
+                    >
+                        Show all activity
+                    </Button>
```
kimpers: If you use `to` instead of `href` it will be a smooth page transition rather than a full reload. Also could you define the path in `WebsitePaths` in `ts/types` for consistency.
```suggestion
                    <Button
                        to={WebsitePaths.AccountActivity}
                        isWithArrow={true}
                        isAccentColor={true}
                        shouldUseAnchorTag={true}
                    >
                        Show all activity
                    </Button>
```

ts/pages/account/activity.tsx
```diff
@@ -8,84 +8,124 @@ import { Breadcrumb } from 'ts/components/ui/breadcrumb';
...
+        return {title, subtitle};
+    } else if (event.eventType === 'staked') {
+        const title = 'Staked';
+        const subtitle = `You staked ${formatZrx(event.eventArgs.zrxAmount).formatted} ZRX with pool ${event.eventArgs.poolId}.`;
+        return {title, subtitle};
+    } else if (event.eventType === 'removed_staked') {
+        const title = 'Removed Stake';
+        const subtitle = `You removed ${formatZrx(event.eventArgs.zrxAmount).formatted} ZRX from pool ${event.eventArgs.poolId}.`;
+        return {title, subtitle};
+    } else if (event.eventType === 'moved_stake') {
```
kimpers, line 49: This should be on the API level but I tested this out locally and it seems if I unstake by moveStake to `NILL_POOL_ID` it results in an `Unkown Event` whereas if I move from one pool to another it is correctly classified as `Moved Stake`.

image

kimpers (approved):
Really good job on this! Looks great!

ts/pages/account/account_detail.tsx
alexkroeger: will do!

ts/pages/account/activity.tsx
alexkroeger: Sounds like an API fix. Can you send me the transaction hashes that generated unknown events?

alexkroeger: Ah it's bc I specified in the event parser `removed_staked` instead of `removed_stake` 🤦 . I'll fix it.

ts/pages/account/account_detail.tsx
alexkroeger: nice that works much better 👍

ts/pages/account/activity.tsx
alexkroeger: Fixed!

johnrjj (approved):
🔥  You love to see it!!!!!!! 🔥 

Great job! Keep em' coming 😄

ts/pages/account/activity.tsx
alexkroeger: image

PR #287, opened 2020-04-23
multiple changes to account activity page to optimize for mobile

alexkroeger:
Altered the account page to be better optimized for mobile.

Changes include:
- Text size responsive to window width
- Timestamp split across 2 lines and is non-separable
- Event detail text does not disappear with lower window width
- Transaction column disappears for small windows

After:
image

Before:
image

### DIFF
--- a/ts/pages/account/activity.tsx
+++ b/ts/pages/account/activity.tsx
@@ -19,11 +19,13 @@ import { utils } from 'ts/utils/utils';
 
 import { logUtils } from '@0x/utils';
 import { useAPIClient } from 'ts/hooks/use_api_client';
+import { useWindowDimensions } from 'ts/hooks/use_window_dimensions';
 import { errorReporter } from 'ts/utils/error_reporter';
 
 export interface ActivityProps {}
 
 const columns = ['Event timestamp', 'Event', 'Transaction'];
+const shortWidthColumns = ['Event timestamp', 'Event'];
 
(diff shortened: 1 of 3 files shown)


ts/pages/account/activity.tsx
```diff
@@ -60,6 +62,9 @@ export const AccountActivity: React.FC<ActivityProps> = () => {
     const networkId = useSelector((state: State) => state.networkId);
     const dispatch = useDispatch();
 
+    const dimensions = useWindowDimensions();
```
kimpers, line 65: I'm completely fine with your solution but there is also a screen width metric available in the state.

```javascript
import { ScreenWidths } from 'ts/types';

const screenWidth = useSelector((state: State) => state.screenWidth);
...
screenWidth === ScreenWidths.Sm
```
would be a mobile width check.

kimpers (approved):
Awesome job on this! 🎉

johnrjj (approved):
🔥 
https://www.youtube.com/watch?v=XHYG2crLab4

ts/pages/account/activity.tsx
alexkroeger: thanks for the heads up! I'll change it. No good to introduce new code unnecessarily

alexkroeger: I decided to keep this as is, because this dynamically adjusts to screen size changes as well. The other method appears to be static on page load.

PR #289, opened 2020-04-26
Revert "Docs for sources with zeros added"

johnrjj:
Reverts 0xProject/website#274


johnrjj:
Not necessary, was a caching issue

PR #291, opened 2020-04-28
Added a CSV export option to the activity page, added a loading screen

alexkroeger:
Changes:

- Added an export to CSV for account events
- Added a loading screen once the wallet connects and before data loads

Testing:
- [x] Tested locally

image

### DIFF
--- a/ts/pages/account/activity.tsx
+++ b/ts/pages/account/activity.tsx
@@ -21,7 +21,7 @@ import { utils } from 'ts/utils/utils';
 import { logUtils } from '@0x/utils';
 import { useAPIClient } from 'ts/hooks/use_api_client';
 import { useWindowDimensions } from 'ts/hooks/use_window_dimensions';
-import { exportCSVFile } from 'ts/utils/csv_export_utils';
+import { exportDataToCSVAndDownloadForUser } from 'ts/utils/csv_export_utils';
 import { errorReporter } from 'ts/utils/error_reporter';
 
 export interface ActivityProps {}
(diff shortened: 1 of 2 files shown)


ts/pages/account/activity.tsx
```diff
@@ -145,17 +172,28 @@ export const AccountActivity: React.FC<ActivityProps> = () => {
...
                 >
                     Activity
                 </Heading>
+                {width > 600 ?
+                <ButtonWrapper>
+                    <Button
+                        isWithArrow={true}
+                        isAccentColor={true}
+                        shouldUseAnchorTag={true}
+                        onClick={() => {exportCSVFile(csvHeaders, delegatorHistory, 'staking_activity'); }}
```
johnrjj: Would add some feedback in case the call fails, can be super simple
```tsx
onClick={() => {
  try {
    exportCSVFile(csvHeaders, delegatorHistory, 'staking_activity'); 
  } catch (e) { 
     reportError(e);
     alert('Error exporting CSV. Try again later.');
  }}
}
```

ts/utils/csv_export_utils.ts
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        str += `${line}\r\n`;
+    }
+
+    return str;
+}
+
+export function exportCSVFile(
```
johnrjj: Super tiny nit, feel free to ignore: Fn naming is a bit ambiguous for return type, prefer more explicit fn names

```tsx
export function exportDataToCSVAndDownloadForUser(  // or something (?)
```

ts/pages/account/activity.tsx
alexkroeger: 👍

ts/utils/csv_export_utils.ts
alexkroeger: good nit!

johnrjj (approved):
giphy-5

PR #292, opened 2020-05-01
Feat: Added rewards section to account page

alexkroeger:
Changes:
- Added Rewards section to account page
- Added loading screen to account page
- Added utility to calculate expected rewards by pool
- Changed pool stats type to include additional fields for operator, member stake
- Added an "EpochWithFees" type to accept an epoch object with total fees

Dependencies:
- https://github.com/0xProject/0x-api/pull/202

Testing:
- [x] Tested locally

Loading screen:
image

Rewards section:
image
image

Rewards section mobile view:
image

### DIFF
--- a/ts/pages/account/account_rewards_overview.tsx
+++ b/ts/pages/account/account_rewards_overview.tsx
@@ -35,17 +35,6 @@ export const AccountRewardsOverview: React.StatelessComponent<RewardOverviewProp
                 >
                     <div>
                         <Heading marginBottom="12px">
-                            Lifetime Rewards
-                        </Heading>
-                        {lifetimeRewards} ETH
-                    </div>
-                </Action>
-                <Action
-                    percentWidth={25}
-                    percentWidthMobile={50}
-                >
-                    <div>
-                        <Heading marginBottom="12px">
                             <FlexHeader>
                                 Estimated for this epoch
                                 <InfoTooltip id="epoch-estimated-rewards">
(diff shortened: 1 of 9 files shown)


mintcloud:
Great stuff! A couple of things on copy

- let's use "Accumulated Rewards" instead of "Available Rewards" to stress the fact that you don't have to withdraw them each time
- I would add a tooltip to the "Estimated for this epoch" : “This estimate is expected to fluctuate at the beginning of an epoch, and progressively converge on the final value"

burnsben:
Screen Shot 2020-05-01 at 4 47 05 PM

Apologies as my original designs were a bit jank. Would it be possible to separate these containers out with a space instead of the divider line? could go ~25% width on the first 2, and 50% on the withdraw section

and align the text left

ts/pages/account/account_rewards_overview.tsx
```diff
@@ -0,0 +1,218 @@
...
+};
+
+const Wrap = styled.div`
+    & + & {
+        margin-top: 20px;
+    }
+
+    @media (max-width: 768px) {
+        padding: 20px;
+        /*background: ${colors.backgroundLightGrey};*/
```
johnrjj:
```suggestion
```

johnrjj (approved):
Looks great, thanks for updating the testing data;

Totally understand the decimal-js dep, I've been meaning to do that b/c of the equation. This page's payload won't have a noticeable difference in relative size b/c of the dep either, tbh.

johnrjj:
Just ping if I should deploy this -- or if you want to, I can send the instructions over (if you haven't deployed before)

PR #295, opened 2020-05-08
Update epoch length to 7 days for ZEIP-77

kimpers:
@alexkroeger let's test through this on staging together with your API changes to make sure everything looks good before we transition to 7-day epochs.


johnrjj (approved):

alexkroeger (approved):
This looks good! I looked into the API and there isn't anything to change. I adjusted how the ending block is calculated if you can take a look (https://github.com/0xProject/0x-api/pull/221), but everything is actually pulling from the on-chain parameters--not using hard-coded values.

kimpers:
> This looks good! I looked into the API and there isn't anything to change. I adjusted how the ending block is calculated if you can take a look ([0xProject/0x-api#221](https://github.com/0xProject/0x-api/pull/221)), but everything is actually pulling from the on-chain parameters--not using hard-coded values.

Nice! I'll test through this on staging tomorrow to sanity check but I think we're good to go 👍

kimpers:
I tested through this and everything looks fine. I can merge and deploy this tomorrow when I wake up. Should be shortly after the epoch change 👍

PR #296, opened 2020-05-17
Temporarily subtract 3 days from next epoch start

kimpers:
The current epoch is 7 days, however, the smart contract still returns 10 days due to the timelock for governance changes preventing us from updating it yet. To not confuse we should temporarily subtract 3 days from all `nextEpoch` starting dates until the contract is up to date.

I will follow up with a PR that reverts these changes that we can deploy after the contract state has been updated.


johnrjj (approved):
👍  thanks for patching this :)

PR #297, opened 2020-05-17
Revert "ADDS temporarily subtract 3 days from next epoch start"

kimpers:
This reverts #296, we should merge this after the epoch length variable has been updated in the smart contracts


alexkroeger:
👀

alexkroeger (approved):
Tested locally as well, looks great, thanks again!

PR #298, opened 2020-05-19
Add documentation for updated price endpoints response formats

kimpers:
Updates documentation for 0xProject/0x-api#228


fragosti:
I'm a bit confused. You didn't add a new endpoint right? You are just modifying the existing `/prices` endpoint correct?

kimpers:
> I'm a bit confused. You didn't add a new endpoint right? You are just modifying the existing `/prices` endpoint correct?

Correct. However, the `/swap/v0/price` endpoint was not documented in the file so I added a new entry. Note that I'm not touching the `/prices` endpoint, only `/price`.

The `/meta_transaction/v0/price` was already documented with the parameters that I added in my PR ([website/blob/development/mdx/api/index.mdx#get-meta_transactionv0price](https://github.com/0xProject/website/blob/development/mdx/api/index.mdx#get-meta_transactionv0price)) so no change needed there.

PR #307, opened 2020-06-15
Disable /explore page for now

kimpers:
It's out of date and needs a revamp. Let's disable it until we have time to fix it up.


johnrjj (approved):
giphy-6

PR #309, opened 2020-06-18
Team Page Updates

johnrjj:


kimpers (approved):
👍

NIT: Maybe we should remove Fabio’s picture as well?

johnrjj:
> 👍
> 
> NIT: Maybe we should remove Fabio’s picture as well?

Thanks for the reminder, done!

PR #323, opened 2020-08-06
migrate website to web3-react

### LINKED ISSUE #319: ZRX Portal - migrate website to use web3-react
Currently [ZRX Portal](https://github.com/0xProject/website) relies on custom logic to handle interactions with web3 wallets in both the [Staking portal](https://0x.org/zrx/staking) and the [governance page](https://0x.org/zrx/vote). To be quicker in adding support for new wallets we would like to migrate our custom implementation to using web3-react (https://github.com/NoahZinsmeister/web3-react/) and its simple hook-based interface.

**Requirements:**

* Swap out all custom web3 logic to standardized *web3-react* integrations for the different wallets.
* Web3 connection information should come from the web3-react hook, it should be removed from global Redux state.
* We want to support the same wallets as we do today, with added support for WalletConnect mobile wallets.
    * Injected wallets (Metamask, Coinbase Wallet, Cipher, Trust Wallet, ImToken, BitPie)
    * WalletLink wallets (Coinbase Wallet)
    * WalletConnect wallets (Trust Wallet, Argent, Metamask Mobile, Rainbow)
* We want keep the same UI dialog for connecting wallets.

### Useful information about current implementation
See below PRs for information about how the current implementation is implemented.

* Submenu for opening the connect dialog
    * [0xProject/website/pull/17](https://github.com/0xProject/website/pull/17)
* Dialog for choosing which wallet to connect
    *  [0xProject/website/pull/32](https://github.com/0xProject/website/pull/32)
    * [0xProject/website/pull/42](https://github.com/0xProject/website/pull/42/files)
* Wallet connection and integration with dialog
    * [0xProject/website/pull/49](https://github.com/0xProject/website/pull/49/files)

iamonuwa:
Closes #319


ts/components/staking/wizard/wizard_flow.tsx
```diff
@@ -710,7 +715,7 @@ export const TokenApprovalPane = (props: TokenApprovalPaneProps) => {
             <ButtonWithIcon
                 onClick={async () => {
                     const allowanceBaseUnits =
-                        (providerState.account as AccountReady).zrxAllowanceBaseUnitAmount || new BigNumber(0);
+                        ({ address: account } as AccountReady).zrxAllowanceBaseUnitAmount || new BigNumber(0);
```
johnrjj: `zrxAllowanceBaseUnitAmount` won't exist on `address`, need to port over from the redux store still

ts/hooks/use_web3.ts
```diff
@@ -0,0 +1,68 @@
+import { useWeb3React } from '@web3-react/core';
+import { useEffect, useState } from 'react';
+
+import { injected } from 'ts/utils/connectors';
+
+export function useEdgerConnect(): boolean {
```
johnrjj: nit: `useEdgerConnect` -> `useEagerConnect`

ts/pages/account/activity.tsx
```diff
@@ -81,26 +81,24 @@ export const AccountActivity: React.FC<ActivityProps> = () => {
         dispatcher.updateIsConnectWalletDialogOpen(true);
     }, [dispatch]);
 
-    const account = providerState.account as AccountReady;
-
     const [isFetchingDelegatorData, setIsFetchingDelegatorData] = React.useState<boolean>(false);
     const [delegatorHistory, setDelegatorHistory] = React.useState<StakingAPIDelegatorHistoryItem[] | undefined>(undefined);
 
-    const accountLoaded = account && account.address;
+    const accountLoaded = account && account;
```
johnrjj:
```suggestion
    const accountLoaded = !!account
```

ts/pages/staking/wizard/wizard.tsx
```diff
@@ -56,33 +51,28 @@ const Container = styled.div`
...
     const [allTimeStats, setAllTimeStats] = useState<AllTimeStats | undefined>(undefined);
+    const [zrxBalance, setZrxBalance] = useState<BigNumber | undefined>(undefined);
+    const [zrxBalanceBaseUnitAmount, setZrxBalanceBaseUnitAmount] = useState<BigNumber | undefined>(undefined);
 
-    const stake = useStake(networkId, providerState);
+    const stake = useStake(chainId, { account, connector });
     const allowance = useAllowance();
 
-    const allowanceBaseUnits = (providerState.account as AccountReady).zrxAllowanceBaseUnitAmount || new BigNumber(0);
+    const allowanceBaseUnits = ({ address: account } as AccountReady).zrxAllowanceBaseUnitAmount || new BigNumber(0);
```
johnrjj: I don't think this is correct either, see above comment. Will need to pull `zrxAllowanceBaseUnitAmount` from elsewhere, unless I'm misunderstanding something. `address` is just a string, right?

ts/hooks/use_web3.ts
iamonuwa: I don't understand this

ts/pages/staking/wizard/wizard.tsx
```diff
@@ -132,7 +122,28 @@ export const StakingWizard: React.FC<StakingWizardProps> = props => {
...
+            // tslint:disable-next-line:no-shadowed-variable
+            const zrxBalance = Web3Wrapper.toUnitAmount(zrxBalanceBaseUnitAmount, constants.DECIMAL_PLACES_ZRX);
+
+            setZrxBalance(zrxBalance);
+            setZrxBalanceBaseUnitAmount(zrxBalanceBaseUnitAmount);
+        };
+
+        // tslint:disable-next-line:no-floating-promises
+        loadBalances();
+    }, [account, chainId, connector]);
```
johnrjj: Ah yeah, looks like you're loading it here. So just need to update where you're getting the balance from.

ts/pages/staking/wizard/wizard.tsx
johnrjj: While you're in there, can you remove the unneeded cast now (`... as AccountReady`) -- makes it easy to miss these things

ts/pages/staking/wizard/wizard.tsx
iamonuwa: I don't think I follow, can you please explain?

ts/pages/staking/wizard/wizard.tsx
```diff
@@ -252,34 +254,13 @@ const SetupStaking: React.FC<SetupStakingProps> = ({
...
-    } else {
-        // Otherwise, we're good to show either the MM or Recommendation pane
-        if (poolId) {
-            // Coming from market maker entry
-            wizardFlowStep = WizardSetupSteps.MarketMakerEntry;
-        } else {
-            // Coming from wizard/recommendation entry
-            wizardFlowStep = WizardSetupSteps.RecomendedEntry;
-        }
-    }
```
johnrjj, line 275: Can we keep this logic up here?

I know it's a tempting refactor to make it cleaner, but I prefer the if statements and enums, as they are more explicit about what's going on in the wizard logic to someone reading the code

iamonuwa: Personally, would love to but it breaks the changes. I was sad removing it :joy: . The account shows up as undefined at all time even when the user is logged in

ts/hooks/use_web3.ts
johnrjj: useEdgerConnect is mispelled -- should be useE**a**gerConnect

ts/pages/staking/wizard/wizard.tsx
johnrjj: Oh nothing actionable here. 

Let's just make sure we fix the `zrxAllowanceBaseUnitAmount` on the comment(s) above.

ts/pages/staking/wizard/wizard.tsx
johnrjj: How about we remove the undefined `providerState.account.state` part then

```ts
    let wizardFlowStep: WizardSetupSteps = WizardSetupSteps.Empty;
    if (!zrxBalanceBaseUnitAmount) {
        // TODO(johnrjj) - This should also add a spinner to the ConnectWallet panel
        // e.g. at this point, the wallet is connected but loading stuff via redux/etc
        wizardFlowStep = WizardSetupSteps.ConnectWallet;
    } else if (!zrxBalance || zrxBalance.isLessThan(1)) {
        // No balance...
        wizardFlowStep = WizardSetupSteps.NoZrxInWallet;
    } else {
        // Otherwise, we're good to show either the MM or Recommendation pane
        if (poolId) {
            // Coming from market maker entry
            wizardFlowStep = WizardSetupSteps.MarketMakerEntry;
        } else {
            // Coming from wizard/recommendation entry
            wizardFlowStep = WizardSetupSteps.RecomendedEntry;
        }
    }
```

johnrjj: or port to the new web3-react account, that would probably work

```ts
const { account } = useWeb3React();

// ...

let wizardFlowStep: WizardSetupSteps = WizardSetupSteps.Empty;
    if (account === undefined) {
        wizardFlowStep = WizardSetupSteps.ConnectWallet;
    } else if (!zrxBalanceBaseUnitAmount) {
        // TODO(johnrjj) - This should also add a spinner to the ConnectWallet panel
        // e.g. at this point, the wallet is connected but loading stuff via redux/etc
        wizardFlowStep = WizardSetupSteps.ConnectWallet;
    } else if (!zrxBalance || zrxBalance.isLessThan(1)) {
        // No balance...
        wizardFlowStep = WizardSetupSteps.NoZrxInWallet;
    } else {
        // Otherwise, we're good to show either the MM or Recommendation pane
        if (poolId) {
            // Coming from market maker entry
            wizardFlowStep = WizardSetupSteps.MarketMakerEntry;
        } else {
            // Coming from wizard/recommendation entry
            wizardFlowStep = WizardSetupSteps.RecomendedEntry;
        }
    }
```

ts/hooks/use_web3.ts
iamonuwa: Okay, thanks

ts/pages/staking/wizard/wizard.tsx
iamonuwa: I just added the allowance data also, I think its resolved

ts/pages/staking/wizard/wizard.tsx
iamonuwa: The issue here is that the data is outside the render method will be called once and it gets called before the page is completely loaded

iamonuwa: Both failed at all times

ts/pages/staking/wizard/wizard.tsx
```diff
@@ -56,33 +51,29 @@ const Container = styled.div`
...
+    const apiClient = useAPIClient(chainId);
 
     const [stakingPools, setStakingPools] = useState<PoolWithStats[] | undefined>(undefined);
     const [selectedStakingPools, setSelectedStakingPools] = React.useState<UserStakingChoice[] | undefined>(undefined);
     const [currentEpochStats, setCurrentEpochStats] = useState<Epoch | undefined>(undefined);
     const [nextEpochStats, setNextEpochStats] = useState<Epoch | undefined>(undefined);
     const [allTimeStats, setAllTimeStats] = useState<AllTimeStats | undefined>(undefined);
+    const [zrxBalance, setZrxBalance] = useState<BigNumber | undefined>(undefined);
+    const [zrxBalanceBaseUnitAmount, setZrxBalanceBaseUnitAmount] = useState<BigNumber | undefined>(undefined);
+    const [zrxAllowance, setZrxAllowance] = useState<BigNumber | undefined>(undefined);
```
johnrjj: We need to pull this from redux instead, otherwise we're duplicating the source of truth and it can get out of sync. (e.g. onUpdateAccountBalances gets called and the zrxBalance balance here won't get updated.

ts/hooks/use_stake.ts
```diff
@@ -69,23 +71,27 @@ export const useStake = (networkId: ChainId, providerState: ProviderState): UseS
...
-        );
-        setStakingProxyContract(
-            new StakingProxyContract(_contractAddresses.stakingProxy, providerState.provider, {
-                from: _ownerAddress,
-            }),
-        );
-        setContractAddresses(_contractAddresses);
-    }, [providerState, networkId]);
+        const loadContract = async () => {
+            const provider = await connector.getProvider();
```
kimpers: This errors out when I try to go through the staking wizard flow after logging into MetaMask.
image. Probably need to trigger this on connector change and make sure that it exists before proceeding.

iamonuwa: Resolved. Thanks @kimpers

ts/components/dialogs/connect_wallet_dialog.tsx
```diff
@@ -133,260 +147,148 @@ const WalletCategoryStyling = styled.div`
...
-    }, [
-        connectToWallet,
-        isMobile,
-        onCloseDialog,
-        providerState.account.state,
-        providerState.displayName,
-        providerState.provider,
-        providerState.providerType,
-    ]);
+        window.location.reload();
```
kimpers: Why are we doing a full page reload here? I can't imagine that being necessary.

ts/components/staking/header/sub_menu.tsx
```diff
@@ -178,11 +134,91 @@ const ConnectButton = styled(Button).attrs({
...
+
+    const logout = async () => {
+        if (typeof window !== undefined) {
+            window.localStorage.removeItem('__WalletLink__:https://www.walletlink.org:Addresses');
+            window.localStorage.removeItem('__WalletLink__:https://www.walletlink.org:SessionId');
+            window.localStorage.removeItem('-walletlink:https://www.walletlink.org:session:id');
+            window.localStorage.removeItem('-walletlink:https://www.walletlink.org:session:secret');
+            window.localStorage.removeItem('-walletlink:https://www.walletlink.org:session:linked');
+            window.localStorage.removeItem('WALLET_CONNECTOR');
+            window.location.reload();
```
kimpers: I don't think there's any need for a page reload here. Calling `deactivate` on web3-react should be enough

ts/utils/providers/provider_state_factory.ts
```diff
@@ -1,89 +1,89 @@
...
+//         const providerState: ProviderState = {
+//             name: 'Fallback',
+//             displayName: walletDisplayName || utils.getProviderDisplayName(provider),
+//             provider,
+//             web3Wrapper: new Web3Wrapper(provider),
+//             account: constants.NO_ACCOUNT,
+//         };
+//         return providerState;
+//     },
+// };
```
kimpers: Feel free to delete these commented out files

ts/connectors/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    rpc: { 1: RPC_URLS[1] },
+    bridge: 'https://bridge.walletconnect.org',
+    qrcode: true,
+    pollingInterval: POLLING_INTERVAL,
+});
+
+export const walletlink = new WalletLinkConnector({
+    url: RPC_URLS[1],
+    appName: '0x',
+    appLogoUrl: '',
```
kimpers:
```suggestion
    appLogoUrl: 'https://0x.org/images/0x_logo.png',
```

ts/redux/async_dispatcher.ts
```diff
@@ -5,30 +5,30 @@ import { Web3Wrapper } from '@0x/web3-wrapper';
...
         networkId: Network,
         shouldAttemptUnlock: boolean = false,
         shouldSetToLoading: boolean = false,
+        connector: any,
     ) => {
-        const web3Wrapper = providerState.web3Wrapper;
-        const provider = providerState.provider;
-        if (shouldSetToLoading && providerState.account.state !== AccountState.Loading) {
+        const provider = await connector.getProvider();
+        const web3Wrapper = new Web3Wrapper(provider);
```
kimpers: Would this not need to be wrapped in the Metamask sub-provider if it's an injected connector?

ts/redux/async_dispatcher.ts
```diff
@@ -49,23 +49,25 @@ export const asyncDispatcher = {
...
+        connector: any,
         dispatcher: Dispatcher,
-        networkId: Network,
+        chainId: number,
     ) => {
         try {
-            const provider = web3Wrapper.getProvider();
-            const contractAddresses = getContractAddressesForChainOrThrow(networkId as number);
+            const provider = await connector.getProvider();
+            const web3Wrapper = new Web3Wrapper(provider);
```
kimpers: Same question as above

ts/utils/configs.ts
```diff
@@ -12,11 +12,11 @@ export const configs = {
     AMOUNT_DISPLAY_PRECSION: 5,
     BACKEND_BASE_PROD_URL: 'https://website.api.0x.org',
     BACKEND_BASE_STAGING_URL: 'https://staging-website-api.0x.org',
-    BACKEND_BASE_DEV_URL: 'https://localhost:3001',
+    BACKEND_BASE_DEV_URL: 'https://staging-website-api.0x.org',
```
kimpers: Could you set these to `https://website.api.0x.org`? `https://staging-website-api.0x.org` does not exist anymore 😳

ts/types.ts
```diff
@@ -76,15 +76,18 @@ export interface AccountNotReady {
...
 export type Account = AccountReady | AccountNotReady;
 
 export interface ProviderState {
-    name: string;
-    displayName: string;
-    providerType?: Providers;
-    provider: ZeroExProvider;
-    web3Wrapper: Web3Wrapper;
-    account: Account;
+    connector?: any;
```
kimpers, line 94: Is this even used anymore? Could we just delete it?

package.json
```diff
@@ -54,6 +60,7 @@
         "deep-equal": "^1.0.1",
         "ethereum-types": "^3.0.0",
         "ethereumjs-util": "^5.1.1",
+        "ethers": "^5.0.9",
```
kimpers: Is this used anywhere if not then let's delete it

package.json
```diff
@@ -98,6 +105,7 @@
         "semver-sort": "0.0.4",
         "styled-components": "^5.0.0",
         "thenby": "^1.2.3",
+        "tiny-invariant": "^1.1.0",
```
kimpers: This also does not appear to be used anywhere, can we delete it?

kimpers (changes requested):

ts/pages/staking/wizard/wizard.tsx
```diff
@@ -132,7 +124,16 @@ export const StakingWizard: React.FC<StakingWizardProps> = props => {
 
         // tslint:disable-next-line:no-floating-promises
         fetchAndSetStakingStats();
-    }, [networkId, apiClient]);
+    }, [chainId, apiClient]);
+
+    useEffect(() => {
+        const loadBalances = async () => {
+            await asyncDispatcher.fetchAccountBalanceAndDispatchToStoreAsync(account, connector, dispatcher, chainId);
+        };
```
kimpers: I think `account`, `connector` & `chainId` can all be null here. I think you need to check that they are all set before fetching the balance

ts/utils/providers/provider_factory.ts
```diff
@@ -1,72 +1,72 @@
...
+//         const injectedProviderIfExists = (window as any).ethereum;
+//         if (injectedProviderIfExists !== undefined) {
+//             return injectedProviderIfExists;
+//             // const provider = providerUtils.standardizeOrThrow(injectedProviderIfExists);
+//             // return provider;
+//         }
+//         const injectedWeb3IfExists = (window as any).web3;
+//         if (injectedWeb3IfExists !== undefined && injectedWeb3IfExists.currentProvider !== undefined) {
+//             const currentProvider = injectedWeb3IfExists.currentProvider;
+//             const provider = providerUtils.standardizeOrThrow(currentProvider);
```
kimpers: I think you need to use this to standardize the injected web3-react provider before it can interact with the staking contracts. The contract wrappers expected a certain structure on the provider

ts/utils/providers/provider_factory.ts
```diff
@@ -1,72 +1,72 @@
...
+//         // HACK(kimpers): TEMPORARY HACK until we revamp wallet handling
+//         // Currently can't use standardizeOrThrow because it's checking this.accounts.length to
+//         // determine if we are already connected which does not exist in WalletLink
+//         const standardizedProvider: ZeroExProvider = {
+//             isMetaMask: false,
+//             isParity: false,
+//             stop: _.noop.bind(_),
+//             enable: (walletLinkProvider.enable as any).bind(walletLinkProvider),
+//             sendAsync: (walletLinkProvider.sendAsync as any).bind(walletLinkProvider),
+//         };
```
kimpers: I think you need to do this hack for the WalletLink and WalletConnect web3-react providers before they can be passed into the staking contract wrappers. Same reason as above.

kimpers (changes requested):

ts/components/dialogs/connect_wallet_dialog.tsx
iamonuwa: Using localStorage to keep the changes instead of state.

ts/redux/async_dispatcher.ts
iamonuwa: Not sure i follow here

ts/components/staking/header/sub_menu.tsx
iamonuwa: Deactivate in most cases alone doesn't do the job.

ts/redux/async_dispatcher.ts
kimpers: I think Web3Wrapper requires MetaMask to be wrapped in a `MetamaskSubprovider` from `@0x/subproviders`

```
const web3Provider = new Web3Wrapper(connector === injected ? new MetamaskSubprovider(provider) : provider)
```

ts/components/dialogs/connect_wallet_dialog.tsx
kimpers: This creates a pretty bad user experience. I would challenge that we need to refresh the entire page, if so then we should not be reading directly from local storage but rather keep it in state (and persist changes to localstorage)

ts/components/staking/header/sub_menu.tsx
```diff
@@ -167,21 +167,12 @@ export const SubMenu = (props: ISubMenuProps) => {
...
 
     const logout = async () => {
         if (typeof window !== undefined) {
-            window.localStorage.removeItem('__WalletLink__:https://www.walletlink.org:Addresses');
-            window.localStorage.removeItem('__WalletLink__:https://www.walletlink.org:SessionId');
-            window.localStorage.removeItem('-walletlink:https://www.walletlink.org:session:id');
-            window.localStorage.removeItem('-walletlink:https://www.walletlink.org:session:secret');
-            window.localStorage.removeItem('-walletlink:https://www.walletlink.org:session:linked');
-            window.localStorage.removeItem('WALLET_CONNECTOR');
+            window.localStorage.clear();
```
kimpers: I don't think we can just clear the entire localStorage, that is fairly destructive and there might be other important information we want to keep there even if the user logs out of the staking porta.

ts/utils/constants.ts
```diff
@@ -61,27 +61,27 @@ export const constants = {
...
+        [Providers.BITPIE]: 'Bitpie',
     } as { [key in Providers]: string },
     PROVIDER_TYPE_TO_ICON: {
-        [Providers.Metamask]: 'metamask_icon',
-        [Providers.CoinbaseWallet]: 'coinbase_wallet_icon',
-        [Providers.WalletLink]: 'coinbase_wallet_icon',
-        [Providers.WalletConnect]: 'walletconnect',
-        [Providers.TrustWallet]: 'trust_wallet_icon',
-        [Providers.ImToken]: 'im_token_wallet_icon',
-        [Providers.Bitpie]: 'bitpie_wallet_logo',
```
kimpers: Here too, why do we need to rename?

ts/types.ts
```diff
@@ -737,18 +738,18 @@ export interface MaterialUIPosition {
...
-    Mist = 'MIST',
-    CoinbaseWallet = 'COINBASE_WALLET',
-    Cipher = 'CIPHER',
-    TrustWallet = 'TRUST_WALLET',
-    WalletConnect = 'WALLET_CONNECT',
-    WalletLink = 'WALLET_LINK',
-    ImToken = 'IM_TOKEN',
-    Opera = 'OPERA',
-    Bitpie = 'BITPIE',
-    Fallback = 'FALLBACK',
```
kimpers: Why do we need to rename all of these? That seems unnecessary

ts/components/dialogs/connect_wallet_dialog.tsx
```diff
@@ -256,13 +260,7 @@ export const ConnectWalletDialog = () => {
...
-                                            <WalletOption
-                                                key={`install-wallet-${key}`}
-                                                name="Install Metamask"
-                                                href="https://metamask.io"
-                                            />
-                                        );
+                                        return null;
                                     } else {
                                         return null;
                                     }
```
kimpers: This if/else makes no sense now

ts/components/dialogs/connect_wallet_dialog.tsx
```diff
@@ -220,14 +220,18 @@ export const ConnectWalletDialog = () => {
...
             address = provider.accounts[0];
         } else {
             address = provider._addresses ? provider._addresses[0] : provider.selectedAddress;
         }
         if (typeof window !== undefined && address) {
             const connectedData = JSON.stringify({
-                name: option.type,
+                name: providerName.includes('Provider')
+                    ? providerName.replace('Provider', '').toUpperCase()
+                    : providerName.toUpperCase(),
```
kimpers: Can we make a utility function that re-maps the name correctly so we don't have to rename everything?

kimpers (changes requested):

ts/components/dialogs/connect_wallet_dialog.tsx
iamonuwa: I'm removing this. Its not necessary.

package.json
```diff
@@ -36,9 +36,15 @@
...
-        "@types/d3": "^5.7.2",
+        "@reach/dialog": "^0.11.2",
+        "@reach/tabs": "^0.11.2",
+        "@web3-react/abstract-connector": "latest",
+        "@web3-react/core": "latest",
+        "@web3-react/injected-connector": "latest",
+        "@web3-react/network-connector": "latest",
+        "@web3-react/types": "latest",
+        "@web3-react/walletconnect-connector": "latest",
+        "@web3-react/walletlink-connector": "latest",
```
kimpers: Can we use the normal versioning scheme here, not `latest`. Otherwise this will definitely break down the line

iamonuwa:
Closing this in favor of #354

PR #327, opened 2020-08-21
Update protocol fee from 150k to 70k everywhere

feuGeneA:
Did a grep for both 150000 and 150,000, and updated all the hits.  These changes are currently deployed to `dogfood`, which is accessible at https://dogfood.0xproject.com/docs/

### DIFF
--- a/mdx/guides/swap-tokens-with-0x-api.mdx
+++ b/mdx/guides/swap-tokens-with-0x-api.mdx
@@ -257,9 +257,9 @@ import * as qs from 'qs';
 ```
 
 ```
-value for gasPrice of 1 wei:  1000000000000150000
-value for gasPrice of 5 wei:  1000000000000750000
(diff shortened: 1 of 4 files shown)


mdx/guides/swap-tokens-with-0x-api.mdx
```diff
@@ -257,9 +257,9 @@ import * as qs from 'qs';
 ```
 
 ```
-value for gasPrice of 1 wei:  1000000000000150000
-value for gasPrice of 5 wei:  1000000000000750000
-value for gasPrice of 15 wei: 1000000000002250000
+value for gasPrice of 1 wei:  1000000000000070000
+value for gasPrice of 5 wei:  1000000000000140000
```
moodlezoup:
```suggestion
value for gasPrice of 5 wei:  1000000000000350000
```

hysz (approved):
LGTM!

PR #336, opened 2020-09-21
Update docs to use 0x API v1 endpoints

kimpers:
This updates the current documentation and guides to be valid for the 0x API `v1` endpoints and adds a basic migration guide for how to successfully migrate from using `v0` to `v1` endpoints. 

Depends on 0xProject/0x-api#361 for `v1` endpoint parity.

### DIFF
--- a/mdx/api/index.mdx
+++ b/mdx/api/index.mdx
@@ -15,7 +15,7 @@ We offer hosted versions for different Ethereum networks. Requests for more netw
 
 Each 0x HTTP API `path` is versioned independently using URI versioning. The format is: `https://api.0x.org/<path>/<version>/<endpoint>`.
 
-For example, you can request `https://api.0x.org/swap/v0/quote` which represents `v0` of the `quote` endpoint in the `swap` path.
+For example, you can request `https://api.0x.org/swap/v1/quote` which represents `v1` of the `quote` endpoint in the `swap` path.
 URLs not adhering to this format are not supported.
 
 A major version bump occurs whenever a backwards incompatible change occurs to an `endpoint`, in which case every `endpoint` in that `path` will be on the next version.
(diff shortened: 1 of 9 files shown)


mdx/api/index.mdx
```diff
@@ -1024,7 +1031,7 @@ Identical to the response schema for `/swap/v0/quote`, with one exception: the `
         }
     ],
     "estimatedGasTokenRefund": "252480",
-    "allowanceTarget": "0x95e6f48254609a6ee006f7d493c8e5fb97094cef"
+    "allowanceTarget": "0xf740b67da229f2f10bcbd38a7979992fcc71b8eb"
 }
 ```
```
dorothy-zbornak, line 1037: Should we remove the following section in mtxs for now?

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
```diff
@@ -129,7 +129,10 @@ The function signature for the `open` function is defined as so:
 
 ```solidity
 // contracts/SimpleMarginTrading.sol
-function open(ZeroExQuote memory quote)
+function open(
+    ZeroExQuote memory quote,
+    address allowanceTarget
```
dorothy-zbornak: Is `ZeroExQuote` ever defined anywhere? Why wouldn't `allowanceTarget` be part of it?

dorothy-zbornak (approved):
Thanks for doing this.
A lot of the solidity is wrong for v1 but I can take that chore on.

mdx/api/index.mdx
kimpers: Is there any reason to remove it? It's still accurate for v0, and I'd be happy to update it for v1 when that is shipped.

I know I peeked at it every now and then when implementing metatx support so it's a good resource to have around 🤓

mdx/guides/develop-a-margin-trading-smart-contract-with-0x-api.mdx
kimpers: Good point! I don't know why but for some reason I thought `ZeroExQuote` was something defined in the protocol spec, I'll update this object to include the allowance target 👍

Issue #355, opened 2020-10-28
Create onchain vote infrastructure in ZRX Portal 


mintcloud:
As mentioned in 0x’s latest governance [roadmap update](https://blog.0xproject.com/0x-governance-in-2020-and-beyond-9e7703bc1c54), 0x Labs is planning to migrate its governance infrastructure into on chain-binding systems in 2021. One of the most battle-tested smart contract architectures in this category is the one created and used by [Compound](https://compound.finance/docs/governance). The system has been functioning smoothly since its launch, and recently adopted also by [Uniswap](https://github.com/Uniswap/governance). In Q1, we are planning to start customizing Compound’s contracts to support the governance of the 0x protocol.  

In the meantime, we’re looking for someone to help us adapt ZRX Portal to support such a governance flow. More specifically the [0x.org/zrx/vote](http://0x.org/zrx/vote) page will have to pull onchain proposals, and provide a flow for ZRX holders to vote and/or delegate their ZRX to other address. 0x specific contracts will not be available for testing this feature, but we're planning to test it with Compound's or Uniswap's implementations.

**Requirements** 

**Web2**
- create onchain vote ("_proposal_") page components and containers. It will likely be a simplified version of the current ["zeip" page](https://0x.org/zrx/vote/zeip-79).
- if necessary, create abstraction of 'vote page', which will accommodate both existing off-chain votes ("_zeip_" page, [example](https://0x.org/zrx/vote/zeip-79)) and onchain vote ("_proposal_") pages - the one added in this effort. **Both vote types will have to co-exist**.
- blend the two types of page in the [/zrx/vote page](https://0x.org/zrx/vote) list. Users will be able to access them interchangeably and will be sorted chronologically.
- add a label to identify the different statuses of the votes (_Upcoming, Live, Passed, Rejected_), based on status.
- link to the [0x forum](https://forum.0x.org/) in the /zrx/vote page
- (nice to have) label to distinguish votes coming from onchain sources ("_proposal_") and off-chain ("_zeip_").
- create delegation widget (example in Compound and Uniswap) with popup flow to change delegation.
- create leaderboard page, similar to [Compound's](https://compound.finance/governance/leaderboard). TBD where to link it from [main governance page](https://0x.org/zrx/vote), and whether to have a widget embedded there.

**Web3** 
- connect through infura/alchemy (already available in ZRX Portal) to pull onchain proposals from either [Uniswap](https://app.uniswap.org/#/vote) or [Compound](https://compound.finance/governance)
[...]

PR #358, opened 2020-12-10
Treasury bounty

shad-k:
Caters to #355


ts/pages/governance/vote_form.tsx
```diff
@@ -239,6 +249,17 @@ export class VoteForm extends React.Component<Props> {
...
+    private readonly _castVote = async (e: React.FormEvent): Promise<void> => {
+        e.preventDefault();
+        const { zeipId: proposalId, operatedPools, providerState } = this.props;
+        const  proposalIdBigNumber = new BigNumber(proposalId);
+
+        const contract = new ZrxTreasuryContract(GOVERNOR_CONTRACT_ADDRESS.ZRX, providerState.provider);
+
+        const { votePreference } = this.state;
+
+        await contract.castVote(proposalIdBigNumber, votePreference === VoteValue.Yes, operatedPools ? operatedPools.map((pool) => pool.poolId) : []).callAsync();
```
moodlezoup:
```suggestion
        const gasInfo = await backendClient.getGasInfoAsync();
        await contract.castVote(proposalIdBigNumber, votePreference === VoteValue.Yes, operatedPools ? operatedPools.map((pool) => pool.poolId) : []).awaitTransactionSuccessAsync({ from: this.props.selectedAddress, gasPrice: gasInfo. gasPriceInWei});
```

moodlezoup: you can import `backendClient` like so:
```
import { backendClient } from 'ts/utils/backend_client';
```

ts/components/governance/wizard_flow.tsx
```diff
@@ -0,0 +1,287 @@
...
+            Confirm Registration
+          </ConfirmButton>
+          <ChangePoolDialog
+              stakingPools={stakingPools || []}
+              onChangePool={onChangePool}
+              isOpen={shouldOpenPoolsDialog}
+              nextEpochStart={nextEpochStart}
+              onDismiss={() => setOpenPoolsDialog(false)}
+              currentPoolDetails={{
+                poolId: '43',
```
kimpers: This should not be hardcoded correct?

ts/components/governance/wizard_flow.tsx
```diff
@@ -0,0 +1,287 @@
...
+  const daysToNextEpoch = nextEpochMoment.diff(todayMoment, 'days');
+  return (
+    <SuccessContainer>
+      <Header>
+        Your voting power is now registered
+      </Header>
+
+      {/* <InfoHeader>Your tokens are now locked.</InfoHeader>
+
+      <InfoHeader>Additional tip</InfoHeader> */}
```
kimpers: This can be removed I guess?

ts/pages/account/dashboard.tsx
```diff
@@ -671,6 +764,47 @@ export const Account: React.FC<AccountProps> = () => {
...
+                style={{ background: 'rgba(0, 0, 0, 0.75)', zIndex: 30 }}
+                isOpen={shouldOpenStakeDecisionModal}
+                onDismiss={() => setOpenStakeDecisionModal(false)}
+            >
+                <StyledDialogContent>
+                    <Heading color={colors.textDarkPrimary} size={28} asElement="h4">
+                        You are already self delegated
+                    </Heading>
+
+                    {   zrxBalance.isGreaterThan(0) ?
```
kimpers: is this logic correct? If `zrxBalance` is 0, show self delegated view.

ts/pages/governance/modal_vote.tsx
```diff
@@ -188,6 +192,176 @@ export const ModalVote: React.FC<ModalVoteProps> = ({ zeipId, isOpen, onDismiss,
...
+        }
+
+        fetchVotingPower()
+            .then(() => setIsFetchingVotingPowerData(false))
+            .catch(err => {
+                setIsFetchingVotingPowerData(false);
+                logUtils.warn(err);
+                errorReporter.report(err);
+            });
+    }, [account.address]);
```
kimpers: I think this effect is missing some dependencies

ts/pages/governance/treasury.tsx
```diff
@@ -0,0 +1,395 @@
...
+    });
+
+    React.useEffect(() => {
+        // tslint:disable-next-line:no-floating-promises
+        (async () => {
+            // const qThreshold = await contract.quorumThreshold().callAsync();
+            // setQuorumThreshold(qThreshold);
+            const hardCodeQuorumThreshold = new BigNumber('10000000');
+            setQuorumThreshold(hardCodeQuorumThreshold);
+        })();
```
kimpers: Can we just make this a constant?

kimpers (changes requested):
Great job on this massive PR 🙏 

There's a bunch of merge conflicts, most are probably from #362. Can you merge in the latest development, pick your changes in conflicts then run `yarn prettier` and sanity check that we didn't lose anything of value?

After that check for any issues with `yarn lint && yarn eslint` (eslint will show some issues but you can ignore any that are not stemming from your code)

ts/components/governance/treasury_summary.tsx
```diff
@@ -0,0 +1,22 @@
...
+import React from 'react';
+
+import { Heading, Paragraph } from 'ts/components/text';
+
+export const TreasurySummary: React.FC<{description: string}> = ({ description }) => {
+  const tokens = marked.lexer(description);
+  const heading = tokens.find((token: Token) => (token as Tokens.Heading).type === 'heading' && (token as Tokens.Heading).depth === 1);
+  const paragraph = tokens.find((token: Token) => (token as Tokens.Paragraph).type === 'paragraph');
+  let summary = '';
+  // @ts-ignore
```
lyaunzbe: just curious, how this is working + why the ignore is needed.  using the find above, isn't paragraph a singular token of type Tokens.Paragraph? (looking at the types, I would think only Tokens.Link has tokens property).

ts/components/governance/wizard_flow.tsx
```diff
@@ -0,0 +1,287 @@
...
+    const selectedPoolObj = stakingPools.find(s => s.poolId === toPoolId);
+    setSelectedPool(selectedPoolObj);
+    setIsDelegationFlow(true);
+    onNextButtonClick(true, selectedPoolObj, zrxAmount);
+  };
+
+  return (
+    <>
+    {
+      !!userZRXBalance ?
```
lyaunzbe: super small nit, but Is this double negation needed since we know userZRXBalance is always number type?

ts/pages/account/account_voting_power_overview.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+                <Stats>
+                    <StatFigure label="Your Delegation" value={formatZrx(delegation).formatted} isNoBorder={true} shouldShowZrxLabel={true} />
+                </Stats>
+            </Flex>
+        </Wrap>
+    );
+};
+
+export const AccountSelfVotingPowerOverview: React.StatelessComponent<SelfVotingPowerOverviewProps> = ({
```
lyaunzbe: i think StatelessComponent type might be deprecated

ts/pages/governance/register.tsx
```diff
@@ -0,0 +1,235 @@
...
+        apiClient.getStakingEpochsWithFeesAsync(),
+        apiClient.getStakingPoolsAsync(),
+      ]);
+
+      const epochStart = epochsResponse.nextEpoch && new Date(epochsResponse.nextEpoch.epochStart.timestamp);
+      setNextEpochStats(epochsResponse.nextEpoch);
+      setNextEpochStart(epochStart);
+      setStakingPools(poolsResponse.stakingPools);
+    })();
+  }, []);
```
lyaunzbe: i think this might be missing `apiClient` in the dep array

ts/pages/governance/treasury.tsx
```diff
@@ -0,0 +1,395 @@
...
+                endDate,
+                createdTimestamp: moment.unix(createdTimestamp),
+                executionEpochStartDate,
+                executionEpochEndDate,
+                tally,
+                executionTimestamp: executionTimestampMoment,
+            });
+            setProposalsLoaded(true);
+        }
+    }, [data]);
```
lyaunzbe: `quorumThreshold` missed in dep array

lyaunzbe (commented):
Left some comments, nothing breaking, just some nitpicks!

ts/pages/governance/treasury.tsx
shad-k: With the new smart contract integration this would be picked up from the contract. Have left this here as a reminder that this is where we can fetch this. Will add a TODO here for better documentation. Will that work?

ts/components/governance/treasury_summary.tsx
shad-k: Yeah, the types from library are misleading. The paragraph does contain more tokens broken down into text and links each of which have a `.text` property on them that contains the raw text.

ts/pages/account/dashboard.tsx
shad-k: Yes, there is a check before this which checks for if the user has voting power or not. So this logic combined with the previous one eventually means `hasVotingPower && zrxBalance === 0`. This essentially means that the user has some self delegated balance.

ts/components/governance/wizard_flow.tsx
shad-k: Fixed

ts/components/governance/wizard_flow.tsx
shad-k: Removed

ts/pages/governance/modal_vote.tsx
shad-k: Added the missing dependencies.

ts/components/governance/wizard_flow.tsx
shad-k: Now that I see this I don't think the double negation is required.  But one case that needs to be handled here is when the balance is 0. Even then we want to show the input. Added that.

ts/pages/account/account_voting_power_overview.tsx
shad-k: Changed it to `React.FC`

ts/pages/governance/register.tsx
shad-k: Added the missing dependencies

ts/pages/governance/treasury.tsx
shad-k: Added the missing dependencies

shad-k:
@kimpers Ran the code through Prettier and linters. There was only one eslint warning from my code which should be fixed now.

@lyaunzbe and Kim Have also replied to all the other review comments. Please take a look and let me know if you guys find anything else that needs to be fixed.

ts/components/governance/wizard_flow.tsx
```diff
@@ -0,0 +1,292 @@
...
+    const onChangePool = (fromPoolId: string, toPoolId: string) => {
+        const selectedPoolObj = stakingPools.find((s) => s.poolId === toPoolId);
+        setSelectedPool(selectedPoolObj);
+        setIsDelegationFlow(true);
+        onNextButtonClick(true, selectedPoolObj, zrxAmount);
+    };
+
+    return (
+        <>
+            {userZRXBalance || userZRXBalance === 0 ? (
```
kimpers, line 191: Since you are defaulting to `0` wouldn't this always be true, regardless of if the wallet is connected or not?

kimpers (approved):
The code looks good, just one more comment. Didn't test the actual website flows though.

ts/components/governance/wizard_flow.tsx
shad-k: When the wallet is not connected, `userZrxBalance` is `undefined`.

kimpers: Oh, I mixed up `userZRXBalance` and `zrxAmount`. My bad 😅

PR #363, opened 2021-01-24
Update team page

johnrjj:
Adds Danning, David, and Romain to the team page


chriskalani:
Thanks for updating this; We also need to remove Steve, Paul, and Jason

PR #367, opened 2021-02-09
chore: update docs for v4 swap endpoints

kimpers:
Updates to new `allowanceTarget` and `to` address and removes the now deprecated `orders` response field and meta transaction section.


xianny (approved):
death to meta-txs!

PR #376, opened 2021-03-24
fix: point the RFQT link on homepage to the right location

phil-ociraptor:
RFQT link is stale, updating to the proper link


johnrjj (approved):
🙏🏻 thank you!!

PR #416, opened 2021-10-12
feat: EIP1559 style gas price estimation

kimpers:


webpack.config.js
```diff
@@ -147,25 +147,12 @@ module.exports = (_env, argv) => {
...
         config.mode = 'development';
         config.devtool = 'cheap-module-eval-source-map';
-        // SSL certs
-        if (fs.existsSync('./server.cert') && fs.existsSync('./server.key')) {
-            config.devServer.https = {
-                ...config.devServer.https,
-                key: fs.readFileSync('./server.key'),
-                cert: fs.readFileSync('./server.cert'),
-            };
-        }
```
kimpers, line 168: With SSL turned on for local dev the page won't even load in Chrome due to an invalid cert 🙈

package.json
```diff
@@ -17,9 +17,9 @@
         "eslint": "eslint 'ts/**/*.{ts,tsx}'",
         "lint": "tslint --format stylish --project . 'ts/**/*.ts' 'ts/**/*.tsx'",
         "fix": "yarn lint --fix",
-        "pre_push": "yarn lint:prettier && yarn lint && yarn test",
+        "pre_push": "yarn typecheck && yarn lint:prettier && yarn lint && yarn test",
```
kimpers, line 20: actually checking types unveiled a few lingering TS type issues:feelsgood:

johnrjj (approved):
LGTM

webpack.config.js
tobernguyen: Is this because the cert is expired?

tobernguyen: I just checked. I didn't see the `./server.key` or `./server.cert` exist.

kimpers: yeah I don't have it either 😂   Previously you could just ignore the invalid config in Chrome but not anymore. I don't think https for local dev adds much value so might as well remove it 🙈

tobernguyen: Cool. Please let me know if you think you need it in the future, I think I can do something about it.

PR #421, opened 2021-10-26
Remove 0x instant and 0x mesh links from footer

rhinodavid:


phil-ociraptor:
Also discussed removing Mesh

PR #437, opened 2022-02-09
Add docs dropdown in header

jlin27:
Add a dropdown that surfaces both the existing docs and the alpha release in gitbook


ts/components/header.tsx
```diff
@@ -7,6 +7,7 @@ import { Link } from 'ts/components/documentation/shared/link';
 import { Button } from 'ts/components/button';
 // import { DropdownProducts } from 'ts/components/dropdowns/dropdown_products';
 // import { DropdownResources } from 'ts/components/dropdowns/dropdown_resources';
+import { DropdownResources } from 'ts/components/dropdowns/dropdown_docs';
```
kimpers:
```suggestion
import { DropdownDocs } from 'ts/components/dropdowns/dropdown_docs';
```

kimpers (approved):
Just an import that needs changing, otherwise looks good to me 👍

PR #444, opened 2022-02-23
Update staking header

DennisSimon:
Hi there, 

here is a very simple PR to update the wording in the staking header based on feedback from discord.
As the PR is a one-liner i also threw in the removal of an unused import that caused linting to fail.


johnrjj (approved):
Thank you!

PR #446, opened 2022-03-17
Add full support for Vercel and add redirects for now deprecated docs

DennisSimon:
This PR adds and configures support for Vercel using the added `vercel.json` file.
Additionally, references to the old documentation were removed (as it is now deprecated) and common link to old
docs automatically redirect to their counterparts in the new docs using vercel redirects. 

A "live-like" preview can be seen here: https://verceltest.0x.org/

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     },
     "resolutions": {
         "node-hid": "2.1.1",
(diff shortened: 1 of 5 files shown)


sohai (approved):
👍

PR #466, opened 2022-10-25
Replace analytic to explorer link

sohai:


DennisSimon:
Added this change in #467 :) Thanks!

PR #488, opened 2023-04-24
remove deprecated link that is slowing down site

wonge97:
Commented out request to instant.0x.org as it's no longer relevant and is slowing down site load time.

### DIFF
--- a/ts/pages/instant/configurator.tsx
+++ b/ts/pages/instant/configurator.tsx
@@ -59,7 +59,7 @@ export class Configurator extends React.Component {
 <html>
     <head>
         <meta charset="utf-8" />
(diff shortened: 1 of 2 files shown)


johnrjj (approved):
LGTM

ts/pages/instant/configurator.tsx
```diff
@@ -59,7 +59,7 @@ export class Configurator extends React.Component {
 <html>
     <head>
         <meta charset="utf-8" />
-        <script src="https://instant.0x.org/v3/instant.js"></script>
+        // <script src="https://instant.0x.org/v3/instant.js"></script>
```
DennisSimon:
```suggestion
```

You can just remove it altogether. We would need a different style of comment in JSX.

